PR #9, opened 2022-10-16
Separate cloud/local install docs

j-f1:
This clarifies which options are useful for running Actual on your own server/NAS and which are best if you want to host it in the cloud.


rich-howell:
Nice work on this Jed @j-f1 thanks. Merged

Issue #20, opened 2022-10-18
🗄️ [Documentation] - Joint Accounts


rich-howell:
### Description

Examples of how to handle joint accounts where there is a joint account on budget and only one half of the joint account is within the budget sheet. 

### Category

Accounts

rich-howell:
This is now in the documentation https://actualbudget.github.io/docs/Budgeting/jointaccounts

PR #30, opened 2022-10-20
Joint accounts

shall0pass:
I've put together two possible ways to handle a joint accounts.  As always, if you'd like me to add others I'm all ears.

Addresses #20


rich-howell:
Thanks for this @shall0pass merged into main

PR #36, opened 2022-10-21
Using Actual

### LINKED ISSUE #11: 🗄️ [Documentation] - Showing cleared and uncleared bank balances isn't easily discoverable
Creating from: https://github.com/actualbudget/releases/issues/241

The expando icon only shows on mouse over, this should be changed to always show so it's discoverable by new users, and possibly even default to showing the expanded balances to enhance awareness.

------
Can you clarify what you're looking for that's different from this?
image

_Originally posted by @MikeBishop in https://github.com/actualbudget/releases/issues/241#issuecomment-1076146400_

rich-howell:
Working on a section about using actual, what each section of the application does and means, this also resolves #11


shall0pass:
I'm not sure some of the images on the "Interface Overview" and "The Sidebar" pages are highlighting the feature well enough.  The images where the app is "softened" to try to highlight the area on the screen is often too distracting.  It takes my eyes a while to search for what is intended to be highlighted.

rich-howell:
We will need to update this to include https://github.com/actualbudget/actual/pull/459

rich-howell:
I am going to merge this as is as it has lots of good information which we can build upon.

PR #56, opened 2022-11-10
scheduling window of 2 days

shall0pass:
I've added a section to discuss the two day window for scheduling transactions.


rich-howell:
Thanks for this @shall0pass

PR #58, opened 2022-11-27
Add important migration instructions

MatissJanis:
The manual migration IMO is mandatory. Without doing it, the budget will eventually break and be lost.

PR #60, opened 2022-12-02
December Release Notes

shall0pass:
A preview of the release notes for December.

Before merging:
1) Re-title release page to correct date -- done
2) Add any new commits that are merged  before release. -- done


docs/Release-Notes/Release-2022.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Actual Server
+
+No changes
+
+
+## 22.10.25
+
+:::warning
+This release includes a breaking change to the sync component, ensure your budget is [backed up](/Backup-Restore/Backups) before you update to avoid data loss
```
MatissJanis, line 35: Could you merge my PR into yours too? IMO this information is critically important.

https://github.com/actualbudget/docs/pull/58

shall0pass: Done 👍

MatissJanis: Great, thanks!

docs/Release-Notes/Release-2022.md
```diff
@@ -0,0 +1,144 @@
+---
```
jlongster, line 1: Why did we rename this page? It'll break existing URLs right? (probably not a big deal, but looks like you want to move to a single page for notes?)

shall0pass: I thought it might be cleaner to group the pages by year with the most recent release at the top so the sidebar doesn't get cluttered.  I don't know of any other links against the URL except for the sidebar page.  I'm open to leaving it as is if that's what you'd prefer.

jlongster: I think you're right that we should consolidate them to a single page. Instead of doing it by year though, let's go ahead and keep it all on one page for now. Can you make this just `Release-Notes.md` or something like that?

shall0pass: Not a problem.

PR #61, opened 2022-12-09
Update release notes regarding 22.12.03

shall0pass:
I've update the release notes regarding the recent update.  Here is a summary.

1) Added some notes for the release page regarding the Docker tags and warning about the 22.12.3 and 22.12.8 tags.  
2) Highlighted the icon change in the features list
3) Added the final published docker tag for each of the releases.

### DIFF
--- a/docs/Release-Notes/Release-Notes.md
+++ b/docs/Release-Notes/Release-Notes.md
@@ -3,12 +3,22 @@ title: Releases Notes
 ---
 
 ## 22.12.03
+**Docker tag: 22.12.9**
+
+:::warning
+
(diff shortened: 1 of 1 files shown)


jlongster:
This looks good to me. Initially I was a little confused at why the versions where added under each actual/actual-server section but then I realized it's because we have different versions for them. We really should split out the release notes between `actual` and `actual-server`; that will reduce the confusion a lot imo.

I agree we need to consistent too about the leading `0`. We had to drop it just so docker would publish it. Learning a lot doing this a few times!

PR #62, opened 2022-12-12
nYNAB export

shall0pass:
I've copied the instructions from https://github.com/actualbudget/actual/tree/master/packages/import-ynab5 to be included in the documentation.

I have not tested these instructions and **assume** they are accurate.


shall0pass:
Accidentally left in another commit.

PR #63, opened 2022-12-12
nYnab export instructions

shall0pass:
I've copied the instructions from https://github.com/actualbudget/actual/tree/master/packages/import-ynab5 to be included in the documentation.

I have not tested these instructions and assume they are accurate.


rich-howell:
@shall0pass there is a conflict here, can you take a look and i will merge it in

shall0pass:
@rich-howell I think that resolves it.

PR #65, opened 2022-12-19
Update your-own-machine.md

tjex:
added a step about pointing the browser to local Actual install (localhost:5060)


rich-howell (approved):
Thanks for the addition @tjex

PR #74, opened 2023-01-12
Fix using HTTPS url for localhost

### LINKED ISSUE #73: 🗄️ [Documentation] - Wrong link for accessing Actual in local installation instruction
### Description

Following the super helpful instructions on installing actual-server locally, I got stuck at the last step ("Accessing Actual") because – as I found out in a discord support request from someone who had the same problem – the linked URL's at the bottom of [https://actualbudget.github.io/docs/Installing/Local/your-own-machine](https://actualbudget.github.io/docs/Installing/Local/your-own-machine) say [https://localhost:5006/](https://localhost:5006/) when it should rather be [http://localhost:5006/](http://localhost:5006/) (i.e., it should be "http" rather than "https"). 

I would love to fix this myself, but I do not have sufficient expertise on github to do so, hence I decided to mention it here. 

### Category

Installation & Configuration, Troubleshooting

j-f1:
Fixes #73, and also takes the opportunity to remove a usage of “simply.”

### DIFF
--- a/docs/Installing/Local/your-own-machine.md
+++ b/docs/Installing/Local/your-own-machine.md
@@ -46,6 +46,6 @@ yarn start
 
 ## Accessing Actual
 
-You should now be able to access [https://localhost:5006](https://localhost:5006) in your browser and you will see Actual.
+You should now be able to access [http://localhost:5006](http://localhost:5006) in your browser and you will see Actual.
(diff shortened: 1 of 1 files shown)


rich-howell:
Thanks @j-f1 merging

PR #91, opened 2023-02-03
Add docs for the config file and HTTPS

j-f1:
HTTPS support will only be available after the next release of Actual, so it may make sense to wait to merge this until the release happens.


TechwizEE:
These updates look great!

rich-howell (approved):

PR #92, opened 2023-02-05
Convert user mentions to reference-style links in the release notes

j-f1:
This fixes a couple of typos and makes the source code easier to read. In the future, add the GitHub URLs for new contributors to the reference section at the bottom of the document, and then just use `[username]` to make the link in the main content


j-f1:
thanks github

PR #97, opened 2023-02-06
Update Fly-updating.md- Add info about updating fly.toml before deployment

00apm:
I have a suggestion for the "Updating Actual" doc page. Since fly.toml is in .gitignore pulling changes via git won't update it. Users should also update their fly.toml before deploying on server. 
I added short info in the Fly-updating.md. However I don't know if this info should also be updated elsewhere.

### DIFF
--- a/docs/Installing/fly/Fly-updating.md
+++ b/docs/Installing/fly/Fly-updating.md
@@ -38,7 +38,7 @@ git pull origin master
 ```
 ![](/img/cmd-25.png)
 
-Once that is done, run the deployment command to push your changes to fly.
+Once that is done, check if your fly.toml file is up to date with fly.template.toml and then run the deployment command to push your changes to fly.
 ```cmd
 flyctl deploy.
 ```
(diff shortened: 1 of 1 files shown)


docs/Installing/fly/Fly-updating.md
```diff
@@ -38,7 +38,7 @@ git pull origin master
 ```
 ![](/img/cmd-25.png)
 
-Once that is done, run the deployment command to push your changes to fly.
+Once that is done, check if your fly.toml file is up to date with fly.template.toml and then run the deployment command to push your changes to fly.
```
tjfinlinson, line 41: @00apm 

I am a big fan of documentation and trying to make it easy to follow for the majority of people. Maybe a bit more hand-holding in the steps here would be helpful, especially if the wrong configuration in your toml file will break your access when doing an upgrade.

Potential Suggestions:
Where to find the fly.template.toml file or a reminder about where to find it
How to open it up and read it, and which sections to look at in the template file that are important for successful upgrade
Reminder on how to update your exsisting fly.toml file

rich-howell: I have added some more description to this.

PR #101, opened 2023-02-09
Add note on migration from old version

### LINKED ISSUE #25: how to migrate data from old subscription app to new self-hosted instance
Short version: After some searching, I figured out how to migrate my data to my new self-hosted server. This issue documents how. I'll happily open a PR to update the documentation if desired. The workaround is to log in to https://app.actualbudget.com which includes the 'export data' button.

Longer version: I could not find any docs or issues detailing how to migrate from the subscription service to the new self-hosted version. I found references to new releases of the Actual Electron App that allow exporting, but that version has not yet been released and installing/building is non-trivial (I have not succeeded yet).

The (temporary) workaround, which I could not find mentioned anywhere, is to log in to the old web version (https://app.actualbudget.com) which runs the latest version of the app, and includes the 'export data' button (under 'settings' from the menu by the Filename). This allows exporting the data, which can then be imported into the web app pointing at your self-hosted server instance. (getting the latest version of the Electron app in order to point it to the self-hosted server is another matter).

Perhaps I could find no issues mentioning this because I'm the only one who didn't think to try this approach ;) ... but I thought I'd mention it for completeness. As I said, I will happily PR to add this to the documentation if desired -- maybe not desirable if a new release is coming soon?
This issue can be closed otherwise. I will close it in a few days if no reply.

hpfast:
closes #25 

It can be confusing for users of the old hosted Actual who want to migrate to the new self-hosted version, because the old Desktop app doesn't have the export button.

This PR adds a note to the Migration page about using the Web version which does have the export button, and links to it from the Installation overview.

### DIFF
--- a/docs/Getting-Started/migration/migration-intro.md
+++ b/docs/Getting-Started/migration/migration-intro.md
@@ -6,9 +6,13 @@ title: "Migration Overview"
 [Click here](simple-sync) if you are looking for the steps to migrate from "full sync" to (the more reliable) "simple sync" method.
 :::
 
+
 Keeping your existing transaction history is important. If you already use a different app, you probably want to migrate it over into Actual.
 
 Right now, only YNAB4 is officially supported. However, the [API](/developers/API/) allows anyone to write a custom importer. We will work with the community to help write other importers soon.
+
+## Migration from the old Desktop App
+The Desktop app that was tied to the old hosted instance of ActualBudget does not have the export button which is available in the Web version and in newer versions of the Desktop app (see https://github.com/actualbudget/actual/pull/71). To export your data, log in to the Web version at https://app.actualbudget.com. Here you can export your data using Settings > Export Data. The export can be imported into newer versions of the Web app and Desktop app.
 
 ## Migration from other apps
 
--- a/docs/Getting-Started/migration/migration-intro.md
+++ b/docs/Getting-Started/migration/migration-intro.md
@@ -12,7 +12,11 @@ Keeping your existing transaction history is important. If you already use a dif
 Right now, only YNAB4 is officially supported. However, the [API](/developers/API/) allows anyone to write a custom importer. We will work with the community to help write other importers soon.
 
 ## Migration from the old Desktop App
-The Desktop app that was tied to the old hosted instance of ActualBudget does not have the export button which is available in the Web version and in newer versions of the Desktop app (see https://github.com/actualbudget/actual/pull/71). To export your data, log in to the Web version at https://app.actualbudget.com. Here you can export your data using Settings > Export Data. The export can be imported into newer versions of the Web app and Desktop app.
+:::note
+Are you coming from the original, managed Actual subscription service? That used an older version of the desktop app which doesn't have the export button, but you can log in to the Web version at app.actualbudget.com to export your data in a format suitable for importing in your new instance. See [Backups](https://actualbudget.github.io/docs/Backup-Restore/Backups) and [Restoring](https://actualbudget.github.io/docs/Backup-restore/Restore).
+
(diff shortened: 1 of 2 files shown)


docs/Getting-Started/migration/migration-intro.md
```diff
@@ -6,10 +6,14 @@ title: "Migration Overview"
 [Click here](simple-sync) if you are looking for the steps to migrate from "full sync" to (the more reliable) "simple sync" method.
 :::
 
+
 Keeping your existing transaction history is important. If you already use a different app, you probably want to migrate it over into Actual.
 
 Right now, only YNAB4 is officially supported. However, the [API](/developers/API/) allows anyone to write a custom importer. We will work with the community to help write other importers soon.
 
+## Migration from the old Desktop App
+The Desktop app that was tied to the old hosted instance of ActualBudget does not have the export button which is available in the Web version and in newer versions of the Desktop app (see https://github.com/actualbudget/actual/pull/71). To export your data, log in to the Web version at https://app.actualbudget.com. Here you can export your data using Settings > Export Data. The export can be imported into newer versions of the Web app and Desktop app.
```
rich-howell, line 15: We have documentation for exporting and importing your data, maybe they could be linked? In the relevant places.

https://actualbudget.github.io/docs/Backup-Restore/Backups
https://actualbudget.github.io/docs/Backup-Restore/Restore

docs/Installing/overview.md
```diff
@@ -15,4 +15,6 @@ While running a server can be a complicated endeavor, we’ve tried to make it f
 
 Once you’ve set up your server, you can [configure it](Configuration.md) to change a few of the ways it works.
 
+If you're coming from the old, hosted version of Actual, you may want to [migrate your data](../Getting-Started/migration/migration-intro.md).
```
rich-howell, line 18: I think we need to be more specific on what the old, hosted version is. Maybe "If you're coming from a managed Actual subscription"

rich-howell (changes requested):

docs/Getting-Started/migration/migration-intro.md
j-f1: I believe it’s also possible to zip up your budget folder (which by default is in `~/Documents/Actual/My-Budget-abc123` on macOS and probably somewhere similar on other platforms)

hpfast:
oops @j-f1 , I 'resolved the conversation' by accident. Good point about being more specific. I'll address that.

docs/Installing/overview.md
hpfast: Good point. I think it will apply specifically to the original app.actualbudget.com, so how about this:

```suggestion
If you're coming from the original, managed Actual subscription service, you may want to [migrate your data](../Getting-Started/migration/migration-intro.md).
```

docs/Getting-Started/migration/migration-intro.md
hpfast: Thanks. I see that the backup-restore docs already draw attention to the fact that some export functionality is only available in the Web version. Perhaps, the following:

```suggestion
:::note
Are you coming from the original, managed Actual subscription service? That used an older version of the desktop app which doesn't have the export button, but you can log in to the Web version at app.actualbudget.com to export your data in a format suitable for importing in your new instance. See [Backups](https://actualbudget.github.io/docs/Backup-Restore/Backups) and [Restoring](https://actualbudget.github.io/docs/Backup-restore/Restore).

Instead of creating an export from the Web app, you can also create a zip archive of your budget folder, which is equivalent. The folder is located at `~/Documents/Actual/My-Budget-abc123` by default on MacOS and Linux. The resulting zip file can be imported into Actual via the Web app.
```

Windows default location ... ?

rich-howell:
Thanks for the changes on this @hpfast I am going to merge this in it's current state and we can work out the detail later.

PR #103, opened 2023-02-10
Removing some of the non Actual based installation methods

rich-howell:
Removing some of the non Actual based installation methods

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -64,21 +64,7 @@ const sidebars = {
               label: 'On Your Own Server',
               collapsible: false,
               items: [
-                'Installing/Docker',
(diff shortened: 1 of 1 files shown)


MatissJanis (approved):

j-f1:
Can you delete the lines entirely along with their corresponding files? We can always get them back from the Git history.

rich-howell:
> Can you delete the lines entirely along with their corresponding files? We can always get them back from the Git history.

They were there for a reason ;) I use Netlify preview to see how things look. Once approved I will and was going to remove which I will do now.

j-f1:
Can you delete the corresponding files too?

PR #105, opened 2023-02-24
Versioning strategy

MatissJanis:
Added a paragraph about the new versioning strategy we plan to adopt starting from `v23.3.x`.

Let me know your thoughts.


trevdor (approved):
I like it. 
And if the last named release of a month ends up delaying into the next month, I say let’s just roll with it.

j-f1 (approved):

PR #106, opened 2023-02-28
Add Easypanel as a deployment method

deiucanta:


rich-howell:
Hello,

Thank you for this pull request, we are trying to keep the documentation limited to deployment methods we support as a team, if you would like to publish this on a blog or somewhere else that is totally fine, we will happily link to it with the understanding that them instructions are managed and maintained by yourself. 

I hope you understand 
Rich

Supernova3339:
ouch

deiucanta:
Hi @rich-howell, thanks for your response.

I wonder if that's the case for Fly.io and PikaPods?

PR #109, opened 2023-03-04
using actual (images and instructions update)

tjex:
Addresses #85 

Updated images and instructions in the 'Using Actual' section. 
Mostly all changes based on the new sidebar layout. 

Also did some file keeping (making folders for images and deleted old images)

The small screenshots still render (unnecessarily) large, I'm not sure how to address that in markdown. 
Happy to update if needed.

PR #110, opened 2023-03-04
Add an image handling component

j-f1:
This PR adds the Docusaurus image plugin, which adds functionality like processing images (which we don’t heavily use at this point), creating a blurred thumbnail for display before the image loads, and computing image size at build time so the page can be laid out properly right away.

I’ve also written a Remark transform to convert ``-style images to the format needed to get the metadata to work correctly.

Finally, I’ve added support for Retina screenshots — if you name an image with `@2x` it will automatically have its preferred size halved to ensure it shows up right.

### DIFF
--- a/docs/Budgeting/returnsandreimbursements.md
+++ b/docs/Budgeting/returnsandreimbursements.md
@@ -51,7 +51,9 @@ Ultimately this is a personal decision.  If you are capable, there is less short
 
 ![](/img/reimbursement-1.png)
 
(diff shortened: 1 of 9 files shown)


rich-howell (approved):

tjex:
Getting some height warping issues with the images now.

Notice how in the below screenshots that the height is warped for one but not the other. 
I expect this is because of retina vs non-retina images as the screenshots I recently did for the docs (#109) are warped as well, and they are from a retina screen.
While the other image below could potentially be from a non-retina. Looks like they're from your screen @rich-howell . 

This behaviour is the same on my localhost and on the public docs wesbite.

Screen Shot 2023-03-04 at 19 41 51

Screen Shot 2023-03-04 at 19 42 28

j-f1:
Fixed that in your PR

PR #118, opened 2023-03-05
Update HTTPS certs when using config.json

JabariHolder:
Based on the code, the paths are no longer `keyPath` and `certPath`. This errors when using the config.json file.

The server looks for `key` and `cert`:
```
key: parseHTTPSConfig(config.https.key),
cert: parseHTTPSConfig(config.https.cert)
```

https://github.com/actualbudget/actual-server/blob/master/src/app.js#L66-L67


j-f1 (approved):
You’re right, thanks for catching it!

PR #121, opened 2023-03-07
[Experimental] Report budget docs

shall0pass:
I'm not a user of this feature, but I thought I could get the ball rolling on adding some documentation.  I'm not 100% sure this is how this budget type is used, so there are likely wrong assumptions.  I'm hopeful someone will either correct me or take the reigns.  There have been several users looking for the features of this budget type, so I think it deserves a little time to explain how it's intended to be used.

**Preview link**: https://deploy-preview-121--zingy-bonbon-b1b286.netlify.app/Advanced/Experimental-Features/report-budget


shall0pass:
Does anyone know when the "Projected Savings" turns into "Saved"?  Does this happen after the month has closed?

image

j-f1:
The heading shows “projected savings” if the month being displayed is after the current month and “overspent/saved” otherwise.

shall0pass:
So March will switch to "Saved" on April 1?  I'll revisit the section where I wrote on this.  I'm not sure I represented that correctly.

shall0pass:
I've marked this as "Ready to Review".  If there are any other sections that need to be added for features/workflows to mention, let me know.

j-f1 (approved):
Looks good to me!

PR #122, opened 2023-03-08
Update Docker docs

j-f1:
- Clarify that the different image sources (Docker Hub / GH Packages) are equivalent
- Don’t endorse using `:edge` as strongly
- Update Docker Compose instructions (see also https://github.com/actualbudget/actual-server/pull/150)
- Remove references to commits/files that were not merged to `master`
- Add instructions for updating the `docker-compose` powered container
- Remove language about asking for the server URL since this will no longer be true (because of this it may make sense to defer merging until the next release comes out)
- Shorten the nginx section since people are unlikely to use it, hide the config file behind a `<details>`


rich-howell:
Applying `do not merge` to avoid this being merged before 23.3.10

j-f1:
I think this is good to go now.

MatissJanis (approved):
Rubber stamp. I have very little experience with Docker

Issue #120, opened 2023-03-06
🔧 [CHANGE] - Image storage and access system


tjex:
### Description

I have an idea that could make the documentation process smoother and faster.

At the moment there is a partial system emerging of a `specific folder + docs section` happening. 
ie images for the `backup-restore` section are stored in a folder at `img/Backup-Restore`.

However there are many cases where the same images are called for throughout different pages in the documentation. 
eg: "click on settings in the sidebar" -> shows an image of the settings button in the sidebar. 

With the current system, each contributor would take a screenshot themselves of the settings button in the sidebar, and store it in the respective folder. 
This creates extra work for the contributor and also creates the need for many more image changes when a change to the UI is done. The repo will also grow in size as more features get implemented as more image are needed. etc.

I think a mixture of this strategy and the the following idea could be better.

For basic menu items, such as buttons in the sidebar, budget view, reports and other such default view that are not changed by specific user input, we could have a folder with associated images within them.

eg: folder = `img/sidebar`, `img/budget-view`, etc

In the sidebar view image file names would be in the direction of: 

`sidebar-settings.png` (showing the 'settings' button in the sidebar)
<img width="231" alt="Screen Shot 2023-03-04 at 13 45 18" src="https://user-images.githubusercontent.com/83029642/223066978-7e278b16-ddd1-4257-9a6d-dcbede870154.png">

`sidebar-accounts.png` (showing the 'accounts' button in the sidebar)
<img width="234" alt="Screen Shot 2023-03-04 at 11 51 52" src="https://user-images.githubusercontent.com/83029642/223067087-ee3f9df3-c457-47cf-845f-37e04bf92dae.png">

For all other images that are needed for that specific page, the contributor/s can use the specific folder for that page. 
eg, instructions for creating rules will be specific to the written instructions in that page, and therefore highly unlikely to be reusable elsewhere in the site. It makes sense then to have them in their own specific folder as it makes managing them easier when working on the page.

### Expected behavior

When writing documentation, contributors can first look in the respective folders for the images that they need for basic settings / functionality references. If it isn't there, they can make one and also put it in that folder, making it available then for others in the future.

I think what could come from this is: 

- less work for contributors and less friction to get involved (as there's less screenshots to process on their end)
- less work to update the 'settings button in the sidebar` when it changes due to UI update. Just changing that file, and giving it the same name, will update all calls to it throughout the documentation
[...]

PR #123, opened 2023-03-08
update readme and begin image restructuring

tjex:
Refers to #120 

Added some instructions to the readme, which include personal design considerations (highlighting with the transparency box and orange/yellow colours otherwise to work better with Actual's color scheme) and a draft of communicating the structure ideas outline below.
Please contest any of it.

Also updated about the `@2x.png` scheme from #110  

Was just thinking to show this preliminary work to check it's in an agreeable direction or not before continuing. 

The changes in this PR boil down to (and looking for feedback on):

1. locking in the idea of reusable images placed in their own subdirectorie: eg `static/img/elements/sidebar/`
2. aligning on some design considerations (highlight tactic / color scheme) and how to communicate them in the readme
3. aligning on a strategy to organise other images (elaborated on below)

For point 3 there are two options I think. 

3a. the images for the page at `github.io/docs/Budgeting/filters`, would be stored in `static/img/filters`
3b. the images for the same page, would be stored in `static/img/budgeting`

Option 3a - ie. one folder per page at `static/img/`:

- will end up creating many more folders in the `img` directory.
- allows the page itself to be moved to a different section of the webpage, and it's image paths to keep making sense.
- helps to maintain mental context as to which images are related to which pages, making it much easier to work with when writing the page
- allows for minimal image naming conflicts
- care will need to be given to name the folder and .md file the same, to keep things understandable later down the track
- if a page gets deleted, deleting its images folder will break any pages that also use those images (but this new structural idea is designed with the idea of that not needing to happen often, if at all. ie a folder for commonly reusable images `static/img/elements/...).

Option 3b -  ie. one folder per subsection at `static/img/`:
- images can be more easily manually searched through in a folder (as all images for all pages under `Budgeting` would be in this folder)
- less thought needs to be given to saving and managing images
- higher likelihood of ending up with redundant images left in the folder after page deletion / change
- if a page moves to a different subsection, either image paths in the `page.md` will need to be updated and images moved to another subsection, which can potentially break other linkings in pages from the same subsection that use the same images. Or the image paths stay unchanged, breaking the relationship between page position in repo structure and where the images are found.


README.md
```diff
@@ -95,9 +97,39 @@ Sub headings are used to split the document into meaningful "chapters" these are
...
+When using images, ensure that the image is large and clear, if the image contains multiple items and buttons, highlight the area you are talking about (as shown below) or create arrows on the image to point to the part you are discussing. Where possible images should be saved as `PNG` format.
+
+The clearest way to highlight areas on a screenshot is the 'transparency' box.   
+
+On Mac:   
+- open the screenshot in preview
+- click the annotate icon
+- from the 'shape' selector, select the 'transparency' tool
+
+![](/static/img/repo/highlighting.png)
```
j-f1, line 114: Files intended to be viewed on GitHub should use relative paths:

```suggestion
![](./static/img/repo/highlighting.png)
```

README.md
```diff
@@ -95,9 +97,39 @@ Sub headings are used to split the document into meaningful "chapters" these are
...
+
+If this doesn't suffice (perhaps the area is too small for the box to be noticable), use an arrow or a colored box. 
+For consistencies sake, use a shade between *orange and yellow*. Other colors either do not work well with the purple colours of Actual (eg blue or green) and red is danger or in our case, a negative balance, double danger!
```
j-f1, line 117: Do you want to pick a color here and write down the hex code so we can stay fully consistent?

Also a small nit:
```suggestion
For consistency’s sake, use a shade between *orange and yellow*. Other colors either do not work well with the purple colours of Actual (eg blue or green) and red is danger or in our case, a negative balance, double danger!
```

j-f1 (commented):
I like this approach! A couple of comments below plus one more: how do you feel about encouraging people to use the built-in test budget for screenshots? I don’t know whether it would be consistent enough in places like the sidebar because it has some inherent randomness so maybe as an alternative you/we could make a demo budget and save it to this repo for testing purposes?

sidebars.js
```diff
@@ -32,7 +32,7 @@ const sidebars = {
         {
           type: 'link',
           label: 'Issue Board',
-          href: 'https://github.com/orgs/actualbudget/projects/1',
+          href: 'https://github.com/actualbudget/actual/issues',
```
rich-howell, line 35: This was for something else, this isn't the Issue Board, which has now been deleted, can you rename this to Issues or something as it isn't the "Issue Board"

README.md
rich-howell: If we are using images in the readme can these come out of the /docs/ folder altogether please if possible.

README.md
```diff
@@ -95,9 +97,39 @@ Sub headings are used to split the document into meaningful "chapters" these are
...
-`### When Using Images`
+### When Using Images
 
-When using images, ensure that the image is large and clear, if the image contains multiple items and buttons, highlight the area you are talking about or create arrows on the image to point to the part you are discussing. Where possible images should be saved as `PNG` format.
+> Screenshots taken on a retina screen should be titled as such: image-name@2x.png
+> This allows Docusaurus to rescale the images where appropriate. 
+
+When using images, ensure that the image is large and clear, if the image contains multiple items and buttons, highlight the area you are talking about (as shown below) or create arrows on the image to point to the part you are discussing. Where possible images should be saved as `PNG` format.
+
+The clearest way to highlight areas on a screenshot is the 'transparency' box.
```
rich-howell, line 107: This is only available on mac - This should read "The clearest way to highlight areas on a screenshot is a 'transparency' box."

README.md
```diff
@@ -95,9 +97,39 @@ Sub headings are used to split the document into meaningful "chapters" these are
...
 
-When using images, ensure that the image is large and clear, if the image contains multiple items and buttons, highlight the area you are talking about or create arrows on the image to point to the part you are discussing. Where possible images should be saved as `PNG` format.
+> Screenshots taken on a retina screen should be titled as such: image-name@2x.png
+> This allows Docusaurus to rescale the images where appropriate. 
+
+When using images, ensure that the image is large and clear, if the image contains multiple items and buttons, highlight the area you are talking about (as shown below) or create arrows on the image to point to the part you are discussing. Where possible images should be saved as `PNG` format.
+
+The clearest way to highlight areas on a screenshot is the 'transparency' box.   
+
+On Mac:
```
rich-howell, line 109: Can we remove this please, else we will need to add in documentation for all platforms. Maybe generalise it a bit more across platforms.

README.md
tjex: Will do 👍 
Although most programs will just have a color pallet to pick from and won't necessarily take hex/rgb code input. 
Will try find the best way for this.

tjex:
Thanks both for the detailed feedback!

> alternative you/we could make a demo budget and save it to this repo for testing purposes?
I was also considering this idea and agree. It would reduce friction for people to get involved if they can just import a demo budget with transactions, accounts, etc already set up. 

It will also create some continuity throughout the docs due to reusing the same account names / payees / etc. 

I can make an issue for it.

README.md
tjex: and @j-f1, regarding the nit: I should've been clearer. Consistency is the name of the devil who peeks over my shoulder every time I make an incongruous colour choice. So for his/her sake, it would be good to do it... jks. Thanks for pointing it out haha.

j-f1 (approved):
Nice! I like this approach. Happy to merge now or wait if you still want to do more in this PR.

tjex:
> Nice! I like this approach. Happy to merge now or wait if you still want to do more in this PR.

great. well I'm aiming to reorganise every image that's so far being used. 
I should have some good time in the coming week to do this. So will just ping you when either it's all done, or I have to drop it for a bit. In which case it would be good to merge, so that others contributing won't create more images to restructure.

tjex:
hey @j-f1 . Smashed it all out now 🤓

There were a bunch of unused images leftover. I moved them into `img/unused`. Nothing broke, but didn't want to just delete as you guys have a better handle on what images may still be in need / etc. So will let someone else deal with that.

But yeah, looking cleaner now and easier to keep a track of :~)
So I'd say you're good to merge.

docs/Installing/fly/Fly-updating.md
```diff
@@ -8,23 +8,23 @@ title: 'Updating Actual'
 Failure to do this may result in budget loss.
 :::
 
-This section focuses on updating the actual-server
+This section focuses on  the actual-server
```
j-f1, line 11: Accidental change?

j-f1:
I believe `actual.png` and `favicon.png` are used to render the website header and favicon and should be moved back to their old locations but the other unused files can be deleted (we can always get them back from git history if needed).

tjex:
ah oops. thought it was just the .ico and .svg. 
done.

docs/Installing/fly/Fly-updating.md
tjex: double oops..

tjex:
spotted one other mistake now too. will have another look through and re-request.

tjex:
now should be good @j-f1

j-f1 (approved):
Thanks for doing this!

PR #127, opened 2023-03-11
[WIP] New Fly install doc

j-f1:


rich-howell:
This draft removes so much useful information, it is one if the most popular guides for getting started with actual. All the helpful screenshots etc are gone. This is a backwards step in my opinion.

j-f1:
Agreed. I want to at least:

- Integrate the existing guides back in
- Add a bunch of screenshots

j-f1:
Updated!

- I took new screenshots for macOS showing how to open and use Terminal
- I integrated the onboarding instructions into the page, and updated them to match our new streamlined flow.
- I added in the FAQ, and added a new one for switching to `edge`
- I removed the guidance on deploying with Git because I believe needing to install Git is too complicated, and focusing on image deploys results in the best experience without us having to worry about things like merge conflicts.

j-f1:
I’m hopeful that this addresses your concerns, but if there are still things you don’t like I would be happy to update!

j-f1:
I see this section of the current Fly docs:

<details>

> Each of these methods come with tradeoffs.
>
> The git approach allows users a high degree of control. With sufficient knowledge, you can target
> the current release, work in progress, or a fork (a unique code history that uses Actual as its
> starting point). You can make fine grained changes to any aspect of your Actual server instance.
> However, some users will find that this level of control adds complexity to their deployment. As
> your deployment is based on your local copy of the source code, you will be responsible for keeping
> track of updates and ensuring your local copy is up to date.
>
> The container image approach offers less granular control. Images are intended to be a specific
> version of the application, identical everywhere that image is deployed. While there are some
> deployment parameters you can alter, you'll only have access to configuration that the current
> version of the application makes available. In return for surrendering source code level access, you
> get relative simplicity. The image includes everything the application needs, and thus none of that
> tooling needs to be installed on your machine. New versions of the image are built by the main
> repository[¹](#deployment_1); if you opt to deploy the `:latest` tag, you can be confident you're up
> to date with the latest full release.
>
> As a (terrible) analogy, the git approach provides you a car with no restrictions. If you decide the
> engine should be tuned differently, you can open the hood and go to work with your wrench. The image
> approach also provides you a car, but the engine bay is sealed. If you want different cylinder
> timing, you'll need a new image.

</details>

IMO given those tradeoffs, and with the availability of experimental features, edge builds, and the potential for third parties to build/publish their own images, images are the best option for Fly.io in all cases. If someone wants to make changes to Actual, they’re best off running it locally. This also means we have fewer deployment methods to worry about.

PR #134, opened 2023-03-16
Update howitworks.md

btehan:
Line 53, where it says That money will appear in next month's **To Budget**. This isn't really happening.  Should this part of the documentation be changed?

### DIFF
--- a/docs/Budgeting/howitworks.md
+++ b/docs/Budgeting/howitworks.md
@@ -50,7 +50,7 @@ Most likely, a common workflow will be to "hold" income you make this month for 
 
 3.  Press `Enter` to hold all available money, or enter a custom amount to hold.
 
-This doesn't do anything except takes out money from the **To Budget** amount for that month, allowing you to "zero it out" and mark it to be used for later. That money will appear in next month's **To Budget**. You can also just leave the **To Budget** amount alone.
+This doesn't do anything except takes out money from the **To Budget** amount for that month, allowing you to "zero it out" and mark it to be used for later. That money will appear in next month's **To Budget**. (insert more accurate description here?) You can also just leave the **To Budget** amount alone.
 
 It is possible to hold money multiple months ahead. If you do this, those dollars will be delay to be budgeted until the last month of the hold. If you need to use that money in the current month to cover over spending, use the **Reset Next Months Buffer** button to bring that money to the current month. This needs to be done for each month there is a hold.
 
--- a/docs/Budgeting/howitworks.md
+++ b/docs/Budgeting/howitworks.md
@@ -50,7 +50,11 @@ Most likely, a common workflow will be to "hold" income you make this month for 
 
 3.  Press `Enter` to hold all available money, or enter a custom amount to hold.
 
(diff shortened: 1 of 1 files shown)


btehan:
After more thought, I'm thinking these are all possibilities:

1. The documentation is inaccurate and needs changes
2. There is a bug where holding a to budget for next month isn't working correctly
3. Most likely I'm doing something wrong

I would appreciate someone else's input on this, please.

shall0pass:
@btehan , when you add income to your budget, the income is added to the "To Budget" category (lack of a better term).  If you're viewing a multi-month view, the "To Budget" amount will propagate forward to all future months.  So if you have 1000 for "To Budget" in this month, the "To Budget" amounts for next month, two months from now, and three months from now will all show 1000.   

The "Hold For Next Month" button really only marks the "To Budget" amount as 'not available for use in this month' and doesn't really move money.  I don't believe there is a bug.

Does this make sense?

btehan:
@shall0pass 
Your explanation makes sense, thank you.  My thought is then that the docs should be adjusted to better reflect your explanation.  But I just don't understand this stuff well enough to make a better explanation.

shall0pass (commented):
How about this?

This removes the entered amount from the **To Budget** availability for the current month.  That money will continue to be available in next month's **To Budget** allowance. Alternatively, you can leave the **To Budget** amount as a positive value, as it will be available for the current and all future months until it is applied to a category.

btehan:
Yeah, I think that's more clear.  Thank you.

btehan:
I've been thinking about this some more and I think part of my confusion is in the contextual menu where it says "Hold for next month".  IDK, maybe I should have opened this up against the file that has that wording in it.  Cause if this option has no effect on the budget for next month, why is it worded like that?  Maybe "hold for next month" could/should be changed?  Food for thought..

shall0pass:
A lot of people have used ynab in the past, and part of the ynab philosophy is to give every dollar a job as soon as it enters the budget.  Actual as a lot of similarities to ynab, especially ynab4.  So instead of giving those dollars a job in a category this month, the "hold for next month" button is moving that chore to next month.  It's sort of less confusing if you look at it in a single month view where you can't see all of the "To Budget" amounts across future months.  I think this feature is a lot nicer than how nYnab does it.  In nYnab, you have to create a special category for next month and then release the category funds at the beginning of the month to start budgeting.  Even ynab4 was better than nYnab for budgeting this month from last months income and Actual's implementation is more flexible than ynab4's.

Kidglove57:
I would echo what @shall0pass has said. I use this feature every month. I posted some possible wording on Discord that would resonate with ex YNAB 4 users. I am copying below for completeness but of course the best suggestions would come from those who have not used envelope budgeting before:
**Suggestion**

rich-howell:
> I would echo what @shall0pass has said. I use this feature every month. I posted some possible wording on Discord that would resonate with ex YNAB 4 users. I am copying below for completeness but of course the best suggestions would come from those who have not used envelope budgeting before:
> 
> **Suggestion** Using Hold for Next Month will ensure that the funds are no longer available to budget in the current month but can still be allocated to the budget in any subsequent month. This will be particularly useful for those who are looking to stop living _paycheck to paycheck_ and instead gradually get one month ahead ie living on last month’s income rather than this month’s.

I have added this :) Thanks @Kidglove57

PR #136, opened 2023-03-20
Improved documentation arround connecting your bank

Soulplayer:
I made a small addition to the documentation page about connecting your bank.


j-f1 (approved):
Looks good to me. Thank you for contributing my! (I’ll also open an issue since this feels like it should be automatic)

PR #137, opened 2023-03-21
Update Configuration.md with a warning about exposing data and config.json

Soulplayer:


j-f1 (changes requested):

Soulplayer:
I think I allowed the suggestion. It's indeed better.

j-f1 (approved):
Thanks!

PR #138, opened 2023-03-22
fix api example config typo. serverUrl -> serverURL

fry:


j-f1 (approved):
Nice catch! Is the API working well for you otherwise?

PR #141, opened 2023-03-26
edge tag contains latest server and client

aharbis:
Based on testing and discussion in Discord, the `edge` tag does contain the latest `desktop-client` as well. Removes an outdated statement in the Docker install documentation that stated otherwise.


j-f1 (approved):
Thanks, had been meaning to do this after making the change but hadn’t gotten around to it!

PR #143, opened 2023-03-30
Update Building-Windows.md

ostat:
reminder to run bash as admin, with detail for the expected exception if you don't


j-f1 (approved):
Thanks for adding this!

PR #147, opened 2023-04-05
Additions to keyboard shortcuts

youngcw:
I added some non documented keyboard shortcuts for the transaction ledger.  I also standardized the capitalization of the existing shortcuts.

I cant decide if its better to have the single letters upper or lower case.  I don't want to imply that a `Shift` press is needed.  Also, in `Ctrl + o` shortcut, the  `o` was lower case maybe to try and not look like the number.  I don't think I have a preference and can change it to upper case if that is preferred.


j-f1 (approved):

j-f1:
Thank you!

PR #150, opened 2023-04-07
:pencil:  update release instructions

MatissJanis:


docs/Developers/releasing.md
```diff
@@ -13,13 +13,9 @@ Releasing @actual-app/api
...
 ```bash
-yarn build:api
-```
-
-Next, bump the version on package.json. Finally, publish it:
-
-```bash
-cd packages/loot-core
+cd packages/api
+vim package.json # bump the version
```
j-f1, line 17: If you do `yarn plugin import version` you can run `yarn version patch` (or `minor`/`major`)

j-f1 (approved):

docs/Developers/releasing.md
MatissJanis: Fancy. I'm very old-skool :D 

I actually want to get rid of these release steps eventually.. we should just push a tag to github and then everything else should happen in CICD. But one step at a time..

PR #152, opened 2023-04-07
Add percentage based targets to table

shall0pass:
This is an addition when https://github.com/actualbudget/actual/pull/858 is merged.


j-f1 (approved):
Thanks!

PR #153, opened 2023-04-16
:bookmark: (23.4.1) fix rule creation from transaction list

MatissJanis:


docs/Release-Notes.md
```diff
@@ -2,6 +2,64 @@
 title: Release Notes
 ---
 
+#### Features
```
j-f1, line 5: Oops?

j-f1 (commented):
Can you run Prettier? Looks like there’s some extra whitespace there.

docs/Release-Notes.md
MatissJanis: oops

j-f1 (approved):

PR #161, opened 2023-04-25
Update nynab.md

aujkb:
Using the command "curl" resulted in an error, as curl is aliased by Powershell to "Invoke-WebRequest". Replacing this with "curl.exe" fixes this issue as it forces the system to actually run curl.

### DIFF
--- a/docs/Getting-Started/migration/nynab.md
+++ b/docs/Getting-Started/migration/nynab.md
@@ -64,7 +64,11 @@ The API key is only shown once, so make sure you copy it down somewhere! More in
 
 Now open a terminal window / command prompt, and enter:
 
-:::note If you receive an error message like: "Invoke-WebRequest : Cannot bind parameter 'Headers'. Cann...". This is because curl is probably aliased to the "Invoke-WebRequest" Powershell commandlet. Use curl.exe instead of just curl in commands below to fix this. :::
+:::note
+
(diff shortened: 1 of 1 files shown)


docs/Getting-Started/migration/nynab.md
```diff
@@ -64,6 +64,8 @@ The API key is only shown once, so make sure you copy it down somewhere! More in
 
 Now open a terminal window / command prompt, and enter:
 
+:::note If you receive an error message like: "Invoke-WebRequest : Cannot bind parameter 'Headers'. Cann...". This is because curl is probably aliased to the "Invoke-WebRequest" Powershell commandlet. Use curl.exe instead of just curl in commands below to fix this. :::
```
j-f1, line 67: It looks like this got collapsed onto one line, can you split it back out to fix the formatting?

```suggestion
:::note

If you receive an error message like: `Invoke-WebRequest : Cannot bind parameter 'Headers'`, this is because curl is probably aliased to the "Invoke-WebRequest" Powershell commandlet. Use `curl.exe` instead of just `curl` in the commands below to fix this.

:::
```

j-f1 (commented):
Thank you for finding this! There’s a small formatting issue but otherwise this looks good to me.

docs/Getting-Started/migration/nynab.md
aujkb: Sure, I was just following a previous note where it was all on one line also..

PR #169, opened 2023-05-07
Add info on template hold flag

youngcw:
Info about using PR#1004 of base app.


j-f1:
Since this functionality is not present in the latest release, can you add a disclaimer to the top of the page that not all features are necessarily available in stable versions yet? That way people hopefully won't be as confused when they try something out and get an error.

youngcw:
Hows that?

j-f1 (approved):
Thanks for adding that!

PR #185, opened 2023-06-10
[WIP] Reorganize page URLs to match the sidebar

j-f1:
Principles:

- lowercase URLs only
- kebab-case
- generally match sidebar structure

🚨 DO NOT MERGE until we have a redirect plan in place so that links don’t break 🚨


j-f1:
Ok, I think this is ready to merge!

PR #211, opened 2023-06-19
Adding architecture details

Shazib:
Maybe these notes should form part of an existing page, but I wanted to put some notes in about the web worker/websockets nuance and clarify the whole frontend/backend thing as it confused me a bit at first.

Perhaps this page could be a stub for adding more 'technical' details as we go. (stick a 'more coming soon' on the bottom?)

Or maybe it should just be called 'Technical Details'


j-f1 (approved):
Nice!

PR #212, opened 2023-06-20
:pencil:  updating the website tagline, the hook

MatissJanis:
Updated the tagline (the hook) for the website.

For people coming in for the first time - they don't know (and care) about the historical context of open source actual vs closed source. So slightly changed the copy to focus on the truly important things - the features.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# https://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -15,15 +15,14 @@ export default function Hello() {
       <div class={`${classes.main} ${classes.container}`}>
         <h1>The same Actual Budget, but Open Source</h1>
         <p class={classes.heroText}>
-          This website is for the Open Source version of Actual Budget, it is a
-          super fast privacy-focused app for managing your finances. You own
-          your data, and we will sync it across all devices with optional
-          end-to-end encryption.
+          Actual Budget is a super fast privacy-focused app for managing your
+          finances. You own your data and can do with it whatever you want.
+          Featuring multi-device sync with optional end-to-end encryption.
         </p>
         <div>
           <Button to="https://demo.actualbudget.org">Try Our Demo</Button>
           <Button primary to="/docs/install/">
-            Set up hosting
+            Install
           </Button>
         </div>
         <div class={classes.heroImage}>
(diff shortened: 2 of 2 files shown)


.editorconfig
j-f1, line 1: :100:

src/pages/index.js
```diff
@@ -13,17 +13,16 @@ export default function Hello() {
...
       <div class={`${classes.main} ${classes.container}`}>
-        <h1>The same Actual Budget, but Open Source</h1>
+        <h1>Enjoy managing your finances</h1>
         <p class={classes.heroText}>
-          This website is for the Open Source version of Actual Budget, it is a
-          super fast privacy-focused app for managing your finances. You own
-          your data, and we will sync it across all devices with optional
-          end-to-end encryption.
+          Actual Budget is a super fast privacy-focused app for managing your
+          finances. You own your data and can do with it whatever you want.
```
j-f1, line 19: This reads slightly better IMO

```suggestion
          finances. You own your data and can do whatever you want with it.
```

MatissJanis:
Let me know what you think.

IMO this is the most important piece of text we have on the page and thus it deserves extra scrutiny. Please tear this to PR to shreds (but suggest alternatives). Lets come up with the best possible tagline and description.

j-f1 (approved):

j-f1:
I think this is a great overview!

src/pages/index.js
MatissJanis: Agreed! Great to have native-speakers on the team.

YagiOoda:
I may suggest adding "zero-based budget" or "envelope budget". 
Maybe "Actual Budget is a super fast privacy-focused app for managing your budget in a zero-based way."
Or "Actual Budget is a super fast privacy-focused zero-based budgeting app for managing your finances."

Not sure if this should go in the tagline or the description, though.

Kidglove57:
I think we definitely need a reference here to the budgeting methodology. I found Actual back in 2019 solely because I was looking for an alternative to the likes of YNAB, Moneywell or Goodbudget. I was very specific about wanting an envelope budgeting solution.

Having said that I'm not sure that the words "Zero Based" (or similar) are as well known as "Envelope Budgeting" (unless perhaps to ex YNAB users).

I would favour something like the following:

_"Your Finances made simple.

Actual Budget is a super fast and privacy-focused app for managing your finances. It is based around the well proven and much loved Envelope Budgeting methodology.

You own your data and can do whatever you want with it. Actual features multi-device sync, with optional end-to-end encryption"_

rich-howell:
As a quick update, I won't be renewing the .org domain next year, I would begin planning whatever change that may be required before then.

MatissJanis:
Updated:
Screenshot 2023-06-23 at 08 26 51

PR #226, opened 2023-07-06
instructions on manual db updates

migillett:
updated `restore.md` to include instructions on manual db updates


j-f1 (changes requested):
Can you clarify that this should only happen if you use edge releases, and if you’re only using stable releases you should report a bug rather than immediately trying these steps?

j-f1 (approved):
Thanks for writing this up!

PR #231, opened 2023-07-09
Change and Add Create Accounts Images

Kidglove57:
Create Local Account image changed and additional image added


j-f1 (approved):
Looks good to me! Thank you for updating this :)

PR #232, opened 2023-07-10
added section for saving filtered transactions for 23.7.1 update

migillett:


Kidglove57:
You beat me to it! Thank you so much for doing this.

The only suggestion I would make is that the pre-existing Example and Conclusion section (at the end in your rewrite), should instead come before the Saved Filters explanation. The Example and Conclusion relate directly to the previous notes regarding using the Filter tool.

Kidglove57:
I do have one more suggestion of a more general nature. I have always felt that the Filter tool notes were misplaced under the Budgeting heading. How would you feel if they were moved to the "Reports" section of the Docs?

migillett:
Deal! I can make the changes mentioned in your first comment. I would say the filters are separate from reports as the reports are more auto-generated and don't necessarily require user input. But a way to auto-import saved filters into the reports page would be super cool.

migillett:
I did remove the `Conclusion` section though. I don't think it was super necessary to help end-users navigate the tool, but if we want that added back in I can absolutely revert that.

Kidglove57:
Thanks so much! That certainly works for me.

By the way, when I wrote the original notes I forgot the more obvious way of using a "Date" filter - to select a Year or Month from the arrow down, rather than a bespoke date period. However, I am more than happy to do a separate PR for that once your addition has been merged.

I see your point re the Reports section. It's just that I use the saved filter as a form of bespoke tabular report, which is why I felt more comfortable with them in the  Reports section. I could not really see their relevance to the Budget section. Maybe they should be a main rather than sub heading? Or go somewhere alongside the notes on transactions?

Kidglove57:
PS when you refer to importing filters into Reports, we're you referring to more than just the current ability to use a saved filter to refine one of the existing reports?

migillett:
Yeah, the date changes might be a bit out of the scope of this specific ticket, so if you want to make a PR for that, go for it!

ahhh, ok, I see what you mean with the reports comment. I think you're right though - moving the whole filtering section to it's own section would make sense.

And as for the filters in the reports page in the app, I actually have another idea. Maybe we could so something like the AWS console where you can add widgets to make a customized dashboard. That data could pull from saved or even custom filters. Have pie charts of monthly expenses or even breakdowns of vendors of a specific category for x days, line graphs of how much your water bill has been for the past x months, etc.

Kidglove57:
Thanks. I'll do a separate PR when this one is merged.

@j-f1 Any thoughts on @migillett ideas for the reports page? This is a bit outside "my pay grade" (aka I am ignorant!).

migillett:
And above my pay grade too for that matter haha.

j-f1:
This PR looks good! I think it would be good to call out the filter functionality on the reports page somewhere but that could be done in a different PR

> And as for the filters in the reports page in the app, I actually have another idea. Maybe we could so something like the AWS console where you can add widgets to make a customized dashboard. That data could pull from saved or even custom filters. Have pie charts of monthly expenses or even breakdowns of vendors of a specific category for x days, line graphs of how much your water bill has been for the past x months, etc.

That sounds super cool! If you’re interested in working on something like that we would love to discuss it (either in Discord or in an issue).

j-f1 (approved):

migillett:
Yeah sounds great! I'd love to chat about it!

PR #243, opened 2023-07-14
Update Docs re delete Rules

Kidglove57:
Update Docs re mechanics to delete Rules

### DIFF
--- a/docs/budgeting/rules/index.md
+++ b/docs/budgeting/rules/index.md
@@ -87,7 +87,7 @@ You can even manually apply all the actions to the transactions. You need to sel
 
 ### Deleting a rule
 
-To delete a rule, go to File > Manage Rules… to view all the rules and select the ones you want to delete. Click the "Delete # rules" button to delete them.
+To delete a rule (or rules), select the ones you want to delete. Then go down to the bottom right of the window and click the "Delete # rules" button.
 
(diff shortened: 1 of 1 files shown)


j-f1:
Can you add a mention that the rule editor is under More → Rules?

Kidglove57:
Looking at the Rules document I think there are a few other places too, where the "old" Actual is referenced. Including an earlier sentence re accessing and editing Rules, which still refers to the file menu.

I will make any obvious adjustments and update.

The only place I personally use Rules very much,  is in Schedules so it is more than possible that I am not aware of some of the more advanced features. But any updating will be better than none!

Kidglove57:
Updated as discussed. I think I have identified and replaced all the references under Rules to the old desktop "File" menu.

Kidglove57:
As an aside, I have had a quick look through and there are still many screenshots in the Docs that need to be updated as they do not include the "More" dropdown.

j-f1 (approved):
Thanks!

j-f1:
Feel free to hit merge yourself once the PR is approved :)

Kidglove57:
@j-f1 I'm learning - slowly! All I can see is a "Squash and merge" button. Is that the one please?

j-f1:
@Kidglove57 That’s the one!

PR #252, opened 2023-07-23
Rewrite managing credit cards page

kyrias:


shall0pass:
When I was starting out, the other guys *ahem* :) were doing webinars to help explain how to use the software.  I actually sat through one back in 2015. I think that's a thing of the past now but I remember it really helped me wrap my head around these ideas.  I really like what you have here.  I was curious what your though would be on including a screen animation that walks through your scenario to give a visual that goes with your story.  They're a lot easier to produce than a video or a webinar.  I walked through your first page and have copied it below.  This is more of a brainstorm idea on how we can get the most out of these explanations.  If it's decided it's not professional enough, that's ok :)

credit_card_1

kyrias:
> When I was starting out, the other guys _ahem_ :) were doing webinars to help explain how to use the software. I actually sat through one back in 2015. I think that's a thing of the past now but I remember it really helped me wrap my head around these ideas. I really like what you have here.

I appreciate it!  :)

> I was curious what your thought would be on including a screen animation that walks through your scenario to give a visual that goes with your story. They're a lot easier to produce than a video or a webinar. I walked through your first page and have copied it below. This is more of a brainstorm idea on how we can get the most out of these explanations. If it's decided it's not professional enough, that's ok :)

I think that's a good idea, though I don't think I'd be the best person to make them.

I think what would be very cool is if we could write down the actions in some simple DSL and then automatically render it into an animation that renders the Actual UI, since then it would be automatically be kept up-to-date every time the docs are built, and wouldn't require manual re-recording when steps change.

One thing I think would be especially neat with that kind of approach is that the user could then get to either see it animated _or_ be able to manually page through the steps to be able to inspect the screen more closely at every step.

docs/budgeting/credit-cards/carrying-debt.md
```diff
@@ -0,0 +1,9 @@
...
+
+If you have an existing credit card debt that you need to track and gradually pay off then it's very important that any new purchases you make with the credit card are pre-funded following the [Within the Budget](./index.md) strategy as there is no point in paying off 100 extra on the credit card if you will then have to put another -100 on the card to pay for groceries the week after.  Instead you should prioritize making sure that your actual expense categories are funded before allocating money to paying down the debt, which will allow you to ensure that you're paying it off in a sustainable manner without accidentally incurring more of it over time.
+
+The simplest way to track preexisting credit card debt in Actual is to create a **Credit Card** category and set it to **Rollover overspending**.  When you add your existing starting balance and any preexisting transactions to your credit card account you need to ensure that the category of those transactions is set to **Credit Card**.  That way you can at any time look at the **Credit Card** category balance to see the status of your debt preexisting debt, and you can ensure that it's getting smaller over time.
```
shall0pass, line 7: That way you can at any time look at the **Credit Card** category balance to see the status of your ~~debt~~ preexisting debt,

docs/budgeting/credit-cards/carrying-debt.md
```diff
@@ -0,0 +1,9 @@
...
+
+Each month when you do your budgeting, _after_ you've accounted for all other necessary expense categories, you allocate however much you want to pay down on that debt into the **Budgeted** column of the **Credit Card** category.  Then when you go to pay your credit card bill you can easily find out how much you should actually pay by taking the **Credit Card** account balance and turning it positive and then adding the current **Credit Card** category balance to it.  If your account balance is -5500 and your category balance is -4500 then you know that you need to pay off 5500 - 4500 = 1000.  This amount will be greater than what you had put into the **Budgeted** column in case you're simultaneously following the [Within the Budget](./index.md) strategy of pre-funding new credit card purchases, and you can without worries pay off the calculated amount knowing that the money has already been explicitly set aside for this specific bill.
```
shall0pass, line 9: Suggested change (adding commas): 

of pre-funding new credit card purchases, and you can, without worries, pay off the calculated amount

shall0pass:
> I think what would be very cool is if we could write down the actions in some simple DSL and then automatically render it into an animation that renders the Actual UI, since then it would be automatically be kept up-to-date every time the docs are built, and wouldn't require manual re-recording when steps change.
> 
> One thing I think would be especially neat with that kind of approach is that the user could then get to either see it animated _or_ be able to manually page through the steps to be able to inspect the screen more closely at every step.

That does sound awesome!  I think we should explore that more later :)

docs/budgeting/credit-cards/carrying-debt.md
kyrias: Good catch!

docs/budgeting/credit-cards/carrying-debt.md
kyrias: Hmm, if you feel strongly about it I can change it, but it wasn't meant to have much emphasis put on it and looks rather visually busy to me.

shall0pass: I don't feel strongly.  It sounded like a run-on when I read it, but an online grammar check says it's fine. So 🤷:).

shall0pass:
LGTM!

PR #255, opened 2023-07-30
Deemphasize 'Cloud' usage in the Sync page

shall0pass:
I'm trying to deemphasize the term 'cloud' here because I think it implies that it's going somewhere unknown.  It should be clear that it's going to a users installed server, and nowhere else.  Pikapods and fly.io, I think would be classic examples where where the term 'cloud' makes sense but wouldn't make sense if someone had it installed on a local PC, NAS or Raspberry Pi.

I left a lot of the document alone further down because it seems like there may be error messages within the app that use the term 'cloud'.

Maybe add an additional explainer in the section that explains the cloud file?

### DIFF
--- a/docs/getting-started/sync.md
+++ b/docs/getting-started/sync.md
@@ -1,8 +1,8 @@
 # Syncing Across Devices
 
-Actual is a different kind of app. Instead of storing all of your data on the internet by default, it stores all of your data locally. That means it works regardless of your network connection, and you always have direct access to your data.
+Actual is a different kind of app. Instead of storing all of your data on your installed Actual server by default, it stores all of your data locally. That means it works regardless of your network connection, and you always have direct access to your data. Your data never goes to any external servers that you don't specify.  It's your data, and you're in control.
 
-We don't want to throw away the internet though. It's too useful. When your data is stored in the cloud, you can access it from any device easily, and you never have to worry about losing your data in case you drop your laptop or phone into a lake.
+We don't want to throw away the internet though. It's too useful. If you decide to grant your Actual server access to the wider internet, you can access it from any device easily, and you never have to worry about losing your data in case you drop your laptop or phone into a lake.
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/sync.md
```diff
@@ -1,8 +1,8 @@
 # Syncing Across Devices
 
-Actual is a different kind of app. Instead of storing all of your data on the internet by default, it stores all of your data locally. That means it works regardless of your network connection, and you always have direct access to your data.
+Actual is a different kind of app. Instead of storing all of your data on your installed Actual server by default, it stores all of your data locally. That means it works regardless of your network connection, and you always have direct access to your data. Your data never goes to any external servers that you don't specify.  It's your data, and you're in control.
```
kyrias, line 3: I feel like this change isn't really correct.  We _do_ store all your data "on your installed Actual server by default."  If you're opening Actual through an instance of the Actual server then you have to explicitly _opt-out_ if you don't want it stored on the server.

docs/getting-started/sync.md
```diff
@@ -1,8 +1,8 @@
...
 
-We don't want to throw away the internet though. It's too useful. When your data is stored in the cloud, you can access it from any device easily, and you never have to worry about losing your data in case you drop your laptop or phone into a lake.
+We don't want to throw away the internet though. It's too useful. If you decide to grant your Actual server access to the wider internet, you can access it from any device easily, and you never have to worry about losing your data in case you drop your laptop or phone into a lake.
```
kyrias, line 5: This is backwards, it's only if you grant the wider Internet access to the Actual server that you can access it from anywhere.

Though you also don't need to have it publicly available to be able to access it from any of your devices so I'm not sure if I think this phrasing is really right either.

docs/getting-started/sync.md
shall0pass: I'll try again :)

docs/getting-started/sync.md
shall0pass: There are so many different ways that people install Actual, it's hard to get it distilled down to a sentence or two.  I'll try again :)

MatissJanis (approved):

PR #259, opened 2023-08-04
:pencil:  removing some outdated information

MatissJanis:


shall0pass (approved):

sethgillett:
@MatissJanis Does using `curl` to retrieve `budget.json` not work for nYNAB anymore?

sethgillett:
Definitely an improvement to be able to import the file directly into the browser! `curl` retrieval feels a little safer than giving a third party tool read / write access to your YNAB account, but most would understandably not want to bother with the hassle of setting up an API key and interacting directly with the command line.

PR #265, opened 2023-08-28
Fix step to skip if no savings account.

daniel-tp:
I believe this was incorrect, as skipping step 6 would mean there isn't anything put into the savings category.


docs/budgeting/index.md
```diff
@@ -111,6 +111,6 @@ This might sound complex, but it's really not! You can get as detailed as you wa
 6. Budget the leftover **To budget** amount to the **Savings** category
 7. Transfer the leftover amount to your savings account and and categorize the transactions with the **Savings** category
```
kyrias, line 112: While you're at it, could you maybe fix the double "and" in this sentence as well?

kyrias (approved):

MatissJanis (approved):
Thanks!

PR #271, opened 2023-10-29
Create reset password doc

mitsumaui:


rich-howell:
Hey,

I am not sure this was ever supposed to be documented, if an attacker has access to run the commands you more than likely have more problems than resetting the password however it gives everyone (not just the budget owner) access to information on how to reset the budget password.

mitsumaui:
OK no worries if not - I figured if you had this message in the UI it might make sense to document the process - as it implies there is a process, and won't take long for a nefarious user to uncover that. 
For the general user who might need to reset their password one is left wondering how when you advertise the capability. 

https://github.com/actualbudget/actual/blob/master/packages/desktop-client/src/components/manager/subscribe/Login.tsx#L71

MatissJanis:
👋 This is great! I would be happy to merge it if you could solve the CI failures.

If an attacker has gained access to the server - he has access to the data file too. So the attacker can just read it without a password (as long as it's not e2e encrypted). So from that perspective: this is not introducing a new attack vector.

PR #272, opened 2023-10-30
Add install docs for Railway.app

Nfinished:
Add install docs for Railway.app, a service similar to Fly.io but a little more non-developer friendly.

### DIFF
--- a/docs-sidebar.js
+++ b/docs-sidebar.js
@@ -95,7 +95,7 @@ const sidebars = {
               type: 'category',
               label: 'In the Cloud',
               collapsible: false,
-              items: ['install/pikapods', 'install/fly'],
+              items: ['install/pikapods', 'install/fly', 'install/railway'],
             },
           ],
         },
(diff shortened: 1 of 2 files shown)


rich-howell:
Hey,

Thanks for this, my understanding is the project has certain methods of deployment that are in the docs, if you want to create a blog post on how to install Actual on another platform that isn't documented you can do that and it can be linked by you.

The reason for this is nobody in the maintainer team has knowledge of additional deployment methods and it would add extra overhead to manage and maintain documentation they don't know about. 

For the non-developer there is already PikaPods which is supported and documented, PikaPods is a 1 click install.

Rich

MatissJanis:
👋 As Rich already pointed out - we can't really maintain all the different solutions for deploying Actual. We only maintain a small, limited set.

However, if you post the Railway install instructions as a blog post - we'll be happy to link to it from the official docs here: https://actualbudget.org/docs/install/#additional-installation-options

PR #273, opened 2023-11-02
Update ynab4 + 5 importer links in api docs

Marethyu1:
So they point to existing implementations. The current links are broken and Im hoping these links are correct - someone please let me know if this is incorrect, thanks


MatissJanis (approved):
thanks!

PR #277, opened 2023-11-18
Adding some contribution guidelines

Shazib:
Adding some additional comments on making changes that could impact electron, as well as starting some stub pages for db migrations and general contribution advice. 

Hopefully these two pages can be updated by others as we discover more things or can be used as launchpads for some kind of automated checks etc.

### DIFF
--- a/docs/contributing/project-details/migrations.md
+++ b/docs/contributing/project-details/migrations.md
@@ -1,0 +1,11 @@
+# DB Migrations Guide
+
+There are some important considerations to make when adding a feature with a db migration.
+
+* DB Migrations also require publishing a new API version as the migrations also need to be applied there.
+
+* The AQL Schema file may need to be updated to match any table changes.
+
+* You must place your migration file in the `loot-core/migrations` folder, with a strict naming convention.
+
+* The naming convention is as follows: `TIMESTAMP_name.sql`. for example. `1694438752000_add_goal_targets.sql`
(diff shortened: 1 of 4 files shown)


Shazib:
Maybe the migrations page should be a sub-section of the existing database page

docs/contributing/project-details/migrations.md
```diff
@@ -0,0 +1,11 @@
+# DB Migrations Guide
+
+There are some important considerations to make when adding a feature with a db migration.
+
+* DB Migrations also require publishing a new API version as the migrations also need to be applied there.
+
+* The AQL Schema file may need to be updated to match any table changes.
```
youngcw, line 7: We should add the file path here

youngcw: I would say the AQL scheme "likely" needs changed not just "maybe"

docs/contributing/project-details/migrations.md
```diff
@@ -0,0 +1,11 @@
+# DB Migrations Guide
```
youngcw, line 1: I think you should add a line that discourages removing columns and tables.  Then a line that says something about being very deliberate with the migration and to think about if anything would likely be added in the future.  That way we minimize the number of migrations

youngcw (approved):

MatissJanis (approved):

PR #290, opened 2023-12-08
dark mode inclusion and dynamic homepage images

youngcw:
Added a dark mode callout in the feature list.

Working on dynamically switching between dark and light screenshots when the page switches themes.

I pulled images from the VRT tests to use for dark images.  They are pretty close to the existing images, but from a slightly larger screen.


MatissJanis (approved):
Nice!

PR #297, opened 2024-01-01
📝 (reverse proxies) Remove duplicate line, add service port for Traefik

t0rb3n:
While deploying with Traefik I found two small errors in the documentation. 
First is the duplicate line, the second is the missing line so that Traefik knows on which port it should send the traffic to.


MatissJanis (approved):
Thanks!

PR #298, opened 2024-01-04
Update Contributor Section - Side Bar Order

Kidglove57:
The purpose is to move Contributor section to the end of the Docs. Currently it appears too early and creates too much clutter for new users

### DIFF
--- a/docs-sidebar.js
+++ b/docs-sidebar.js
@@ -249,8 +249,12 @@ const sidebars = {
           ],
         },
         'community-repos',
+        ],
+        },
       ],
-    },type: 'category',
+    },
+    {
+      type: 'category',
       label: 'Contributing',
       link: {
         type: 'doc',
--- a/docs-sidebar.js
+++ b/docs-sidebar.js
@@ -29,9 +29,6 @@ const sidebars = {
     'vision',
     'releases',
     {
-      
-    },
-    {
       type: 'category',
       label: 'Getting Started',
       collapsible: false,
@@ -300,6 +297,7 @@ const sidebars = {
           ],
         },
       ],
+    },
   ],
 };
 module.exports = sidebars;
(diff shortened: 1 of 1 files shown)


Kidglove57:
I am unclear as to why the preview has failed

rich-howell:
I am not really sure this PR is suitable, why has the contributing section been moved right to the bottom of the sidebar, nested within the Help & Support category? 

This is how the sidebar looks now with your proposed change.

image
 
I think the contributor section looked fine where it was, what was the confusion that people were running into? Also what clutter did it create? It was a single collapsed header.

Kidglove57:
I agree the Contributing section should not be nested. That is my lack of understanding as to how to use Docasuarus.

The purposes was to get new users (quite a lot of whom may be non technical) straight into setting up and using Actual. 

Contributing then logically comes  a little later in the order of things and of interest only for those with the required technical skills.

MatissJanis:
Fixed the issues so build preview would work.

MatissJanis (approved):
I think moving it below-the-fold makes sense as most folks accessing the page won't be interested in the "contributing" information. They will be, however, interested in: what is this? how do I run a demo for it? once I'm happy with the demo: how do I run it for myself? And then maybe sometime later they _might_ be interested to contribute.

MatissJanis:
For reference how others do this (since I'm literally reading the Electron docs right now..): contribution pages at the very bottom.

Screenshot 2024-01-05 at 19 35 12

rich-howell:
> For reference how others do this (since I'm literally reading the Electron docs right now..): contribution pages at the very bottom.
> 
> 
> 
> Screenshot 2024-01-05 at 19 35 12
> 
> 

Yes but you have put it nested under Help & Support which is wrong.

PR #303, opened 2024-01-13
Adding Create Schedule From Transaction

xentara1:
As discussed..
https://github.com/actualbudget/actual/pull/2222


youngcw:
Ok, one last thing. Would it make more sense to put this in the schedules page instead of the bulk actions page?

xentara1:
I thought that, but you access it via the bulk actions page. Took precedence from the link schedule docs. Happy to move if you think makes more sense in schedules. On Jan 25, 2024, at 17:57, youngcw ***@***.***> wrote:﻿
Ok, one last thing. Would it make more sense to put this in the schedules page instead of the bulk actions page?

youngcw:
I think it makes more sense in the schedules page since most of the time it will be used with a single transaction not multiple

youngcw:
Thanks.  Ill merge this in once we know that the current bug is fixed and the feature is working.

Issue #296, opened 2024-01-01
⭐ [Documentation] - Fly.io


rich-howell:
### Description

I think it would be a good idea to inform people of this https://community.fly.io/t/we-are-going-to-start-charging-for-dedicated-ipv4-in-january-1st/15970 because this comment on the docs page for Actual `If you follow the steps in this guide, you will remain well within the free plan limits. For more details, check out [Fly’s pricing documentation](https://fly.io/docs/about/pricing/)` is no longer true. 

### Documentation URL

https://actualbudget.org/docs/install/fly

tjex:
Although, it does say for new apps, that no unexpected charges will be had.

> New apps have been getting shared IPv4 when first deployed for a while, your new apps should be good to go with no unexpected charges. Also, just creating an app doesn’t assign IPs immediately.

Also on the fly pricing site:

> Each application receives a [shared IPv4 address](https://fly.io/docs/reference/services/#shared-ipv4) and unlimited [Anycast IPv6](https://fly.io/docs/reference/services/#ipv6) addresses for global load balancing.

> Dedicated IPv4 addresses are $2/mo.

As the install/fly doc is for new setups, and it appears like new apps will get a shared ipv4 (free of charge), maybe it doesn't need to change?

tjex:
And actually, my fly instance, made last year, already has a  shared ipv4.
So yeah, it seems like the news is about dedicated ipv4 that will now cost, but that used to be free? 🧐

Or am I missunderstanding / overlooking something?

tjfinlinson:
This probably just need a little bit of clarification on the docs page and making sure it's in alignment with what fly.io says about it.

bruno-cobucci:
As someone who is coming to Actual, I also felt confused when reading the docs. The thing is, there used to be a free plan in Fly.io - that's even written in the docs: 

> If you follow the steps in this guide, you will remain well within the free plan limits.

But, in a recent change, there's no more a free plan. The default, start plan is the "Hobby plan", which costs a minimum of $5/month. See [this discussion](https://community.fly.io/t/i-got-charged-today-for-the-hobby-plan-and-im-confused-as-to-why-cause-fly-io-didnt-send-any-emails/17969/3), for example.

Maybe the docs should be updated to adress this change in Fly.io. Specially considering that this error is in the main "Installing Actual" section.

tjex:
[...]

PR #311, opened 2024-02-06
cautionary notice about new Fly pricing

tjex:
As discussed in [#296](https://github.com/actualbudget/docs/issues/296) the new
pricing structure of Fly is confusing and not being communicated well.

This PR puts up a temporary notice explaining this, so that new users are
informed and won't get slugged with an surprise $5 charge (if that is indeed the
case... 🤷).

It also offers a temporary solution to move ahead with a local install, which
can then be pushed to Fly later on. So new users won't just hit a dead end
either.


MatissJanis (approved):
Thanks!

PR #315, opened 2024-02-10
Goal templates: Edit note that currency symbols other than $ are supported

LarsStegman:
The regex for the parser is implemented to match any currency symbol:

https://github.com/actualbudget/actual/blob/5a67b7e822eb6d10b1e661869b71083e0b4cc8f6/packages/loot-core/src/server/budget/goal-template.pegjs

### DIFF
--- a/docs/experimental/goal-templates.md
+++ b/docs/experimental/goal-templates.md
@@ -44,8 +44,7 @@ You are welcome to have other lines in your note, but the #template line must ma
 
 ### Notes
 
-- $ sign is optional, `#template $50` and `#template 50` are the same.
-- Other currency symbols are supported.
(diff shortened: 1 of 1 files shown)


docs/experimental/goal-templates.md
```diff
@@ -45,7 +45,7 @@ You are welcome to have other lines in your note, but the #template line must ma
 ### Notes
 
 - $ sign is optional, `#template $50` and `#template 50` are the same.
```
youngcw, line 47: Instead can we just change this line and get rid of the line that was changed?  Maybe this could read "Currency symbol is optional"

LarsStegman: Fixed

PR #319, opened 2024-02-23
Add text to reconciliation section about lock icon

shall0pass:


docs/accounts/reconciliation.md
```diff
@@ -26,6 +26,14 @@ When the cleared amount of the Actual account ledger and the value you entered i
 
 ![](/img/reconcile/reconcile-4.png)
 
+Upon pressing done, all cleared transactions will receive a lock icon.  This represents the fact that they were included in a reconciliation and makes it harder to accidentally change.
```
youngcw, line 29: Can we add more about what the lock means? Like "this confirms that the transaction is accurate with your bank statement" or something

shall0pass: I considered this to be covered in the **Workflow** section, line 7.

PR #324, opened 2024-02-28
Update bank-sync.md

Kidglove57:
Explain advantages of setting an opening balance when starting from scratch with Actual and Go Cardless

### DIFF
--- a/docs/advanced/bank-sync.md
+++ b/docs/advanced/bank-sync.md
@@ -101,7 +101,7 @@ GoCardless
 
 ### Frequently Asked Questions
 
-**The best way to start from scratch in Actual with Go Cardless?**
+**Does Actual sync automatically with your Bank?**
 
 At this moment, it is not yet possible for Actual to automatically sync with your bank. You need to do this manually by going to "All Accounts" and pressing "Sync".
 


docs/advanced/bank-sync.md
```diff
@@ -101,8 +101,15 @@ GoCardless
 
 ### Frequently Asked Questions
 
-**Automatic Bank Sync**
+**The best way to start from scratch in Actual with Go Cardless?**
```
youngcw, line 104: Looks like this header was meant to be something like "Does the sync happen automatically" ?.  Is that true? Its currently the same as the other heading.

Kidglove57: Yes - thanks for spotting this! I have no idea how that happened. I have got completely stuck working out how to edit the changes proposed  in a PR like this. And I still can't work it out!

The heading of the first section should have remained as:
"Automatic Bank Sync"

Or as you say it may be better amended to:
"Does Actual sync automatically with your Bank?"

How can I change this without starting g from scratch please?

youngcw: If you edit just that line you should be able to send up a new commit.  So to make the PR you had to make a new commit, so all you need to do this time is make another commit that only changes that one line then push it up again.  Git should handle everything and we should see the one line change here in the PR

Kidglove57: I'm really sorry to have wasted your time I have had a look in GitHub desktop and I have  totally confused myself. I almost started a brand new PR!

This is like a foreign language to me. If necessary, I'll close this PR and we can start from scratch.

youngcw: Its all fine.  Did you try to fix the line?  I still only see the original changes and not the line fix

youngcw:
I fixed it.  Thanks for helping clean things up.  The docs could use some more love

youngcw (approved):

PR #337, opened 2024-03-22
Cleanup documentation for local groups

shall0pass:
I think there is a better example than the one I've proposed here.  I'm not convinced the example I've given would even provide an expected result to the user.  I'd appreciate any feedback of those interested and the PR #2480 could be improved to deliver more expected results.

### DIFF
--- a/docs/experimental/monthly-cleanup.md
+++ b/docs/experimental/monthly-cleanup.md
@@ -105,7 +105,7 @@ Yes.
 
 Method 1:
 
-One way to do this is to have a Utilities holding category with $500 budgeted.  Within that category, use `#cleanup utilities source`.  Within all of the remaining utilities categories (power, gas, water, etc), use `#cleanup utilities`.  This adds the remaining categories to the `utilities` group.  The script will fill any overspending from the holding category.
+One way to do this is to have a Utilities holding category with $500 budgeted.  Within that category, use `#cleanup utilities source` and `#cleanup utilities sink`.  Within all of the remaining utilities categories (power, gas, water, etc), use `#cleanup utilities`.  This adds the remaining categories to the `utilities` group.  The script will fill any overspending from the holding category and return any remaining money to the holding category.
 
 Method 2: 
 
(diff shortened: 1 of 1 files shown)


shall0pass:
The part where I don't think my example gives a preferred outcome is that the 'holding' source category simply redistributes to the sink categories.  Maybe a more preferred outcome is that the 'holding' category is used to fill underfunded categories within the group.  The 'sink' designation could be put into the same category as the 'source' allowing unused money to return to the 'holding' category, but there would need to be a way to specify a category belonging to a group without including it as a sink category.

youngcw:
I think you should add an example that shows a template with a group used.

shall0pass:
> I think you should add an example that shows a template with a group used.

Thanks for suggesting that.  I found a bug that I was able to correct in the process of proofing out my examples.

docs/experimental/monthly-cleanup.md
```diff
@@ -19,21 +19,33 @@ Enable this feature alongside the **Goals** experimental feature by enabling **G
...
 |#cleanup source|This is a source of money to be reused at the end of the month|Electricity is intentionally over budgeted each month and the excess is used to pay down debt|
 |#cleanup sink|This is a category where extra money will be moved. Weight: 1|This can be a vacation, debt, or other savings category where you want to accelerate the savings rate|
 |#cleanup sink 2|This is a category where extra money will be moved. Weight: 2|This can be a vacation, debt, or other savings category where you want to accelerate the savings rate|
 
+### Local group source and sinks
+Local groups can be defined to target certain categories for more refined control. You can have many groups by changing the group name.
+
+|Syntax|Description|Application|
+|---|---|---|
+|#cleanup source _Group_|This is a source of money to be reused at the end of the month with any Group categories|A reimbursement holding category exists for making small loans to family or friends|
```
youngcw, line 36: This example is out of order.  It should be `#cleanup _Group_ source` right?

youngcw (approved):

PR #340, opened 2024-03-26
:fire: remove non-existing api methods from api docs

MatissJanis:
Removed api methods that do not exist. Since they don't exist - they should not be in the docs.

It would be nice to implement them though.. but until they are implemented - lets not mislead people.


twk3 (approved):
:+1:

PR #348, opened 2024-04-14
fix: Open Collective link

huyenltnguyen:
Changes the Open Collective URL from `.org` to `.com` (the `.org` is not accessible).


MatissJanis (approved):
Thanks!

PR #359, opened 2024-05-25
Add a download page to the website

youngcw:
This is a super rough start for a download page for the website.  Please make contributions since this is pretty bad so far.  Can you tell Im not a front end dev? :laughing: 

In order for the links to always work we either need to:
* Modify our electron build to not add versions to the filename
* Add some way for the page to know the current version and use that in the requested filename

For now they are hard coded with the version so they wont work next month unless something changes.

For now the OS logos are colored so that they can be seen in both light and dark modes.  Im sure there is a better way than making them purple.


psybers:
Purple icons probably would work, but could you make it match the other purple text?  Right now I feel it is too bright (when in dark mode).  A darker shade of purple would probably work.

psybers:
Should this page center its body, similar to the other top-level pages?

psybers:
## Current (`#8812e1`):
image
image

## Proposed (`#6B46C1`):
image
image

youngcw:
The color I used is the one from the logo.  It could be dimmed a bit and still match the theme though.  Feel free to tweak this.  It took me too long just to get this far.

psybers:
@youngcw Is there a way to get the download links to always point to the latest version?  To avoid having to update this file every release?

youngcw:
@psybers So I think the setup that would work would be to change the electron build script to not add the version to the filename.  Github already has a URL that always points to the most recent release (using "latest instead of a tag name).  Thats already part of the PR, but the file names themselves need to be right and not change.

psybers:
Another option: could the tooling that generates release notes also somehow update this file? Then the assets could rename versioned.

youngcw:
> Another option: could the tooling that generates release notes also somehow update this file? Then the assets could rename versioned.

I would prefer something like this, where the page can tell what the current version is and know how to point to the right place

RubenOlsen:
I do not know the CSS behind the dark/light theme, but is it possible to have the logos (and links) in even lighter colours when the dark theme is used?

I think the contrast between the background and the logos needs to be bigger. I am unfortunately off for 2 weeks vacation or else I could have had a stab of this one.

src/pages/download.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Desktop Client
+|||
+|:--:|:--:|:--:|:--:|
+|<Winsvg width="100" height="100" fill="#6B46C1"/>|<Macsvg width="100" height="105" fill="#6B46C1"/>|<Linuxsvg width="100" height="100" fill="#6B46C1" />|<Linuxsvg width="100" height="100" fill="#6B46C1"/>|
+|[Windows](https://github.com/actualbudget/actual/releases/latest/download/Actual-windows.exe)|[Mac](https://github.com/actualbudget/actual/releases/latest/download/Actual-mac.dmg)|[Linux (Appimage)](https://github.com/actualbudget/actual/releases/latest/download/Actual-linux.AppImage)|[Linux (Flatpak) ](https://github.com/actualbudget/actual/releases/latest/download/Actual-linux.flatpak)|
+
+
+## Server Download
+See the install guides on how to download and install the server:
```
MatissJanis, line 22: **❓ question:** ‏what's "a server"? A short paragraph about the benefits of using a server (i.e. it allows to sync data across devices) would be good to have here for newcomers.

MatissJanis (approved):
LGTM! Having a download page is better than striving for the ✨ perfect ✨ page, but never actually releasing it.

(just one small nitpick in my other comment)

MatissJanis (approved):

PR #364, opened 2024-06-03
Remove web terminal section from Fly.io docs

matthewlehner:
The web terminal is no longer available on Fly.io and has caused a bit of confusion for folks who are trying to get Actual deployed.

### DIFF
--- a/docs/install/fly.md
+++ b/docs/install/fly.md
@@ -24,28 +24,7 @@ the form. Note that Fly requires that credit card details for sign up. See [thei
 
 ### Accessing the `fly` command line tool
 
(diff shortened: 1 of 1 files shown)


matthewlehner:
Just bumping this PR before I start at-messaging folks!

I work at Fly.io and love that you're recommending us as a place to host Actual (I've started using it myself!) but we had folks in our forums posting stuff like this because they're confused by the web CLI mentions in the directions:

https://community.fly.io/t/guide-to-set-up-actual-budget-as-a-web-only-app-not-on-my-windows-machine-so-my-wife-on-can-also-log-in-from-her-mac/20151/1

psybers (approved):

PR #366, opened 2024-06-05
feat:add deploy to cloud - zeabur

iamshaynez:
I made a zeabur template for Actual Budget deployment. It's an alternative PaaS solution to choose, relatively cost can be less than PikaPods.

Please kindly review.

Thanks!

### DIFF
--- a/docs-sidebar.js
+++ b/docs-sidebar.js
@@ -51,7 +51,7 @@ const sidebars = {
               type: 'category',
               label: 'In the Cloud',
               collapsible: false,
-              items: ['install/pikapods', 'install/fly'],
+              items: ['install/pikapods', 'install/fly', 'install/zeabur'],
             },
           ],
         },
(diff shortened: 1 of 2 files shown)


iamshaynez:
'Zeabur' is band name of a PaaS platform provider, so not necessary a known words. Please reveiw.

iamshaynez:
Should I add 'Zeabur' in expect.txt?

MatissJanis:
Hi! Thanks for adding this, but we (the maintainers) can only maintain a limited set of install/deployment instructions. We don't have expertise in every framework/system/SaaS in the world. So we have picked a few opinionated ones to recommend by default. For all the rest: we recommend creating a blogpost on your personal website (or medium.com or another blogging platform). And then link to the post here: https://actualbudget.org/docs/install/#additional-installation-options

PR #374, opened 2024-06-27
fly.io new pricing details update

wdpk:
fly.io pricing details have changed for new users. new users can only sign up for a pay-as-you-go plan where usage is billed monthly. as this is about deploying actual, it should be aimed at new users, not returning users with maybe-existing legacy plans(like the hobby plan) details about the legacy plans can still be found at the linked page.


MatissJanis (approved):
Thanks!

PR #386, opened 2024-07-03
Created page on the topic - How envelope budgeting works

RubenOlsen:
I have added this page because I think the software used for envelope budgeting should definitively have some documentation on how to actually do envelope budgeting.

### DIFF
--- a/docs/budgeting/envelope-budgeting.md
+++ b/docs/budgeting/envelope-budgeting.md
@@ -1,0 +1,390 @@
+# Envelope budgeting
+
+## Introduction
+Envelope budgeting is a time-tested method of managing money. Traditionally, people would divide their physical cash
+into different physical envelopes, each representing a specific expense category like groceries, entertainment, or savings.
+When all the cash in an envelope was spent, no more money could be used for that category until the next budgeting
+period. This method helped people control their spending by making them aware of their financial limits.
+
+In Actual Budget, we bring this traditional system into the digital age. Instead of physical envelopes,
+Actual Budget uses categories that serve the same purpose. Each category is like a virtual envelope where you
+allocate a portion of your income. This modern approach not only maintains the benefits of the envelope system
+but also addresses some of its limitations.
+
+One key difference is that traditional envelope budgeting often overlooks fixed expenses like rent, mortgage,
+utility bills, and other regular payments. These fixed costs are crucial and must be accounted for to have a
+complete picture of your finances. Actual Budget ensures that every fixed or variable expense is
+considered. This comprehensive approach makes AB much more effective than the traditional cash-stuffing method.
+
+A significant advantage of using Actual Budget is increased awareness of expenses. Knowing where
+every part of your money is a powerful step towards financial control and stability.
+
+
+
+
+## Zero-Based Budgeting
+Zero-Based Budgeting is a key principle in electronic envelope budgeting systems like Actual Budget. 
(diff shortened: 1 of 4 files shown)


docs/budgeting/envelope-budgeting.md
```diff
@@ -0,0 +1,402 @@
...
+Once you have identified the categories for your expenses and incomes, the next step is 
+to decide how much money to allocate to each category. This allocation ensures that 
+every part of your financial life is accounted for and helps you manage your money effectively.
+
+A practical way to determine these amounts is to use the average of at least the last 
+three months' bank statements. Better yet, it is using the previous six months. 
+This average provides a realistic estimate of your spending and income patterns, allowing 
+you to set initial budget values that reflect your financial habits.
+
+> **Tip:** To help you with this task, you can enter previous months' data into Actual Budget.
```
youngcw, line 80: I think you should add on to this tip about that those old transactions will need to be categorized, which can be an annoyance to many.  Use better words than those though.

youngcw (commented):
Hi! This is really good.  Im worried that we are getting a bit redundant between this, the main "Budgeting" page and the "Starting Fresh" page.  Could you look over those and see what we can do to be a bit more cohesive.  Maybe this article could be added to the "Getting Started" section instead of buried down in the technical docs.  Also maybe the Starting Fresh page could point people here if they are unfamiliar with envelope budgeting, since I think they cover the same things, but this is much more in depth.  What do you think?

docs/budgeting/envelope-budgeting.md
RubenOlsen: Thanks for the kind feedback.

I have several ideas on how to make the end-user documentation better, but I wanted to show off my work before putting a lot of stuff on the table :-) 

Also, of course, ask for permission before submitting a PR with a huge number of reorganisations. 

I'll take a stab at the Getting Started section, but I will do this in several steps, which will result in several PRs rather than one big one. 

Regarding rewriting the tip: I agree with you on this, and my plan was/is to have a separate page for this rather than putting a lot of stuff inside the tips. A proper explanation of how to import old transactions demands more than a few lines of text - and I believe that adding a lot of tech stuff to this kind of article makes that information (which will be useful for a lot of people) buried alive. We should remove this tip until that other page has been written.

I also have one other suggestion on annotating images that I would like to change in the readme - and that is that we remove the suggestion on using transparency and only use boxes. For the vision impaired, there is not enough contrast between what's highlighted and what's not when using transparencies. I also suggest we change the colour to pure red for the same reason:

image

Sticking to boxes is much easier to achieve with many of the stock-out-of-the-OS screenshot tools than using the transparency box.

I am also planning to redo several of the screenshots - I noticed several of what we have today is a bit dated compared to how the GUI actually looks.

To sum up - within this PR:
- Move this page to Getting Started.
- Remove the tip until we have a page explaining exactly that.

Next steps:
- Reorganize the Getting Started section into a more beginner-friendly one.

I do care a lot about end-user documentation that are for non-techies - ask the people I lead at work :-D

youngcw: I think all you are proposing is fine.  I think the box thing should be fine, but should be in its own PR.

The docs could use a lot of help to be more consistent, and its true that many of the screenshots are old and should be updated.  So please do try and improve things

youngcw (commented):
Once the spelling issues are fixed, this should be good.

youngcw (approved):

PR #390, opened 2024-07-05
Explain condition types on Rules and Filtering pages 

wdpk:
Add information about "matches" being regex.

~~PR 9/10ths of the way there, still need to update screenshots. Is there a style policy regarding dark/light mode screenshots? It seems like currently it swaps between the two.~~


RubenOlsen:
> Is there a style policy regarding dark/light mode screenshots? It seems like currently it swaps between the two.

This is not mentioned in the README.md file.

My take on this is that for documentation purposes, and from a usability perspective, we should strive to use light mode as much as possible. 

If we want to annotate our screenshots, we should use a clear, high-contrast colour. I think that the proposed colour for annotating boxes does not have enough contrast when seen on a dark background.

What is also very important is that the screenshots we use are clean and readable on a small screen (mobile), so when screenshotting a whole browser window, one should strive to make it as "small as possible" without sacrificing readability.

I am working on the Tour of Actual right now, which I will do only using the light mode and will **not** take screenshots of the full screen browser on my 5K screen :-D

image

wdpk:
i don't use light mode, so i was confused on where some of the screenshots seemed dark. i now see that's a default part of light mode. took new screenshots using light mode. added a border by using fuzzy-select to grab everything that wasn't the box content, deleting it, 'shrinking' the selection by 1px and filling the gap with the color that matches the date selector's edge for contrast against the docs-site-background-color. for the `Categories` screenshot, I have `matches` selected with `.+?` in the screenshot as filled text, as a gentle nudge for someone familiar with regex to make the connection from the screenshot alone, and someone unfamiliar with regex that this box may be more complex than a simple 'match' might seem.

removed the screenshot with notes, because it has the same filter options as categories, so two screenshots can effectively show you all there is without extra.

docs/reports-filters/filters.md
```diff
@@ -27,14 +27,14 @@ Now select **Filter**. You will see from this drop-down that there are a number
 
 The great thing is that you are not limited to just one Filter. You can select multiple filters. In effect stacking one upon another. An example will help illustrate this.
 
-There are further options to choose within each area, in order to narrow the Filter still further. Here are three just to illustrate the choices – Dates, Categories and Notes:
+There are further options to choose within each area, in order to narrow the Filter still further. Here are a couple just to illustrate the choices – Dates, and Categories:
```
RubenOlsen, line 30: Suggestion: _There are further options within each area to narrow the filter further. Here are a couple to illustrate the choices – Dates and Categories:_

This simplifies the first sentence a bit. Also, commas are not used if you have only one "and" in the sentence (I do the same error all the time :-D)

docs/reports-filters/filters.md
```diff
@@ -27,14 +27,14 @@ Now select **Filter**. You will see from this drop-down that there are a number
...
+
+`Matches` is regex, the other condition types are further explained at [Rules Page](../budgeting/rules/index.md).
```
RubenOlsen, line 32: I suggest using _regular expression_ rather than regex. Regex / regexp is very _developer-tribe-talk_. 

Only some people know what regular expressions are, so maybe we should point them to https://regextutorial.org/ or https://en.wikipedia.org/wiki/Regular_expression?

RubenOlsen (changes requested):
In general believe it can be approved as-is without my proposed changes - so I am not going to fight over this :-D

docs/reports-filters/filters.md
wdpk: this makes sense, made it point to regextutorial

docs/reports-filters/filters.md
```diff
@@ -27,14 +27,14 @@ Now select **Filter**. You will see from this drop-down that there are a number
 
 The great thing is that you are not limited to just one Filter. You can select multiple filters. In effect stacking one upon another. An example will help illustrate this.
 
-There are further options to choose within each area, in order to narrow the Filter still further. Here are three just to illustrate the choices – Dates, Categories and Notes:
+There are further options within each area to narrow the filter further. Here are a couple to illustrate the choices – Dates and Categories:
+
+`Matches` is *[regular expressions](https://regextutorial.org/)*, the other condition types are further explained at [Rules Page](../budgeting/rules/index.md).
```
youngcw, line 32: This should link directly to the subsection that you are referring too.  Use ../budgeting/rules/#condition-types for your link

wdpk: changed!

PR #392, opened 2024-07-07
Add reference to the Python library

bvanelli:
Adds the link to the Python library I wrote for actual called [actualpy](https://github.com/bvanelli/actualpy) to the list of community projects. It's also released [on Pip](https://pypi.org/project/actualpy/).

Library implements the core functionality of most of the original API, by a mix of reverse-engineering the UI calls and the checking the source code.

Experimental features cover bank syncs, schedules and rules.


MatissJanis (approved):
Thanks for your contribution!

PR #393, opened 2024-07-07
Extended the PikaPods documentation for non-technical users

RubenOlsen:
I have extended the PikaPods documentation to cater to non-technical users. This should be quite straight forward.


docs/install/pikapods.md
```diff
@@ -2,13 +2,125 @@
...
+![](/img/pikapods-setup/pikapods-7-add-pod-resoruces.png)
+
+
+
+## Your pod is now ready to be used
+
+When click on the green **OPEN POD** you will be taken your Pod. 
+
+The address for your Actual Budget is found in the Domain field. In the screenshot example, this is
+https://berserk-bullmastiff.pikapod.net/budget/. Yours will be something completely different.
```
youngcw, line 104: If this isn't something we want clicked, then we shouldn't make it a link.  Maybe use highlighting with " ` " (grave tick) instead.

RubenOlsen: That was not my intention either. I was not aware of this auto-magic-feature-to-make-unicorns-fly-in-the-sunset. 

Will fix.

youngcw (approved):

Kidglove57:
This is a great bit of work - thank you so much @RubenOlsen

PR #396, opened 2024-07-08
Update docs to be time-neutral by removing outdated references, update related screenshots to where these issues occured

wdpk:
categories and filters pages had references to an outdated release scheme, and updates in a particular year/month combo.

the docs should be time-neutral for released features, so that additional updates after release aren't needed to remove "new as of [xx]". users can follow commits/prs if they are interested in the timeline, but otherwise docs should present current situation.

also on the pages where these were, one made specific reference to outdated screenshots, and one had no screenshots at all, so added screenshots there.

i checked for "update" "release" and "2022/3/4" and found no other outdated references like these two.


docs/budgeting/categories.md
```diff
@@ -58,7 +58,16 @@ You can see in the budget that our categories have now been successfully merged
 ![](/img/categories/CategoriesMerged.png)
 
 ## Expanding, Collapsing or Hiding Categories
-Updates, in June and July 2023, added the ability to (a) *expand or collapse category groups* by clicking on the three dots at the top of the Category list and (b) *hide categories* that are no longer needed (using the down arrow by each category name). Hidden categories can be viewed, when needed, by clicking on the three dots at the top of the category list. The screenshots above do not currently reflect these enhancements.
+
+Category groups can be *expanded or collapsed* by clicking on the three dots at the top of the Category list. Categories that are not currently needed can be *hidden* by using the down arrow by each category name. Hidden categories still impact your budget.
```
RubenOlsen, line 62: _Categories that are not currently needed can be *hidden* by using the down arrow by each category name._ 

I do understand what the text is saying - but for a new user, this sentence makes less sense.

I suggest having 2 paragraphs, one for _expanding / collapsing the category **group**_, and one paragraph dealing with _hiding and unhiding categories within the category **group**_.

Regarding hiding categories, we should have some examples of why one would like to do this. When I teach zero-som budgeting and categories I use "Saving to buy a new washing machine" as an example of a category that should be hidden when the washing machine is bought.

docs/budgeting/categories.md
```diff
@@ -58,7 +58,16 @@ You can see in the budget that our categories have now been successfully merged
...
+
+Hidden categories can be viewed, when needed, by clicking on the three dots at the top of the category list.
```
RubenOlsen, line 64: To simplify the language, I suggest a variant of _When needed, hidden categories can be viewed by clicking on the three dots at the top of the category list. _

RubenOlsen (changes requested):
I have only one real issue, and that is my comment on changes in the categories  regarding hide/unhide/expand/collapse. These are two different use cases and I suggest two paragraphs and not trying to tie all of these into the same paragraph. 

Also a few suggestions on language - nothing major.

docs/budgeting/categories.md
wdpk: i like the idea behind providing an example, can incorporate into this pr

wdpk: could use some help with the grammar i think but

youngcw:
For the images that are calling out a specific button it is nice to the have the called out option hovered.  That highlights so its more visible and obvious.  I think hidden categories and delete category images are the only ones that need fixed.

wdpk:
> For the images that are calling out a specific button it is nice to the have the called out option hovered. That highlights it so its more visible and obvious. I think hidden categories and delete category images are the only ones that need fixed.

got it! will take some new ones.

docs/budgeting/categories.md
```diff
@@ -57,8 +57,17 @@ You can see in the budget that our categories have now been successfully merged
...
+## Managing Categories
+
+Category groups can be *expanded or collapsed* by clicking on the arrow to the left of the Category group name. All category groups can be expanded or collapsed at once through use of the three dot menu at the top of the category list.
+
+Categories that are not currently needed can be *hidden* by using the down arrow by each category name. An example is a budgeting savings for a large cost(e.g. washing machine). After the purchase is made, the category can be hidden. Hidden categories still impact your budget. When needed, hidden categories can be viewed by clicking on the three dots at the top of the category list. 
+
+![](/img/categories/CategoryToggleHidden.png)
+
+Hidden categories, when shown, are displayed in a low-contrast color.
```
psybers, line 68: Lower, not low. Low contrast implies it might be hard to read, which (hopefully) is not the case.

```suggestion
Hidden categories, when shown, are displayed in a lower-contrast color.
```

docs/budgeting/categories.md
```diff
@@ -57,8 +57,17 @@ You can see in the budget that our categories have now been successfully merged
 
 ![](/img/categories/CategoriesMerged.png)
 
-## Expanding, Collapsing or Hiding Categories
-Updates, in June and July 2023, added the ability to (a) *expand or collapse category groups* by clicking on the three dots at the top of the Category list and (b) *hide categories* that are no longer needed (using the down arrow by each category name). Hidden categories can be viewed, when needed, by clicking on the three dots at the top of the category list. The screenshots above do not currently reflect these enhancements.
+## Managing Categories
+
+Category groups can be *expanded or collapsed* by clicking on the arrow to the left of the Category group name. All category groups can be expanded or collapsed at once through use of the three dot menu at the top of the category list.
+
+Categories that are not currently needed can be *hidden* by using the down arrow by each category name. An example is a budgeting savings for a large cost(e.g. washing machine). After the purchase is made, the category can be hidden. Hidden categories still impact your budget. When needed, hidden categories can be viewed by clicking on the three dots at the top of the category list.
```
psybers, line 64:
```suggestion
Categories that are not currently needed can be *hidden* by using the down arrow by each category name. An example is budgeting savings for a large cost (e.g. washing machine). After the purchase is made, the category can be hidden. Hidden categories still impact your budget. When needed, hidden categories can be viewed by clicking on the three dots at the top of the category list. 
```

docs/budgeting/categories.md
```diff
@@ -57,8 +57,17 @@ You can see in the budget that our categories have now been successfully merged
 
 ![](/img/categories/CategoriesMerged.png)
 
-## Expanding, Collapsing or Hiding Categories
-Updates, in June and July 2023, added the ability to (a) *expand or collapse category groups* by clicking on the three dots at the top of the Category list and (b) *hide categories* that are no longer needed (using the down arrow by each category name). Hidden categories can be viewed, when needed, by clicking on the three dots at the top of the category list. The screenshots above do not currently reflect these enhancements.
+## Managing Categories
+
+Category groups can be *expanded or collapsed* by clicking on the arrow to the left of the Category group name. All category groups can be expanded or collapsed at once through use of the three dot menu at the top of the category list.
```
psybers, line 62:
```suggestion
Category groups can be *expanded or collapsed* by clicking on the arrow to the left of the Category group name. All category groups can be expanded or collapsed at once by using the three-dot menu at the top of the category list.
```

psybers (changes requested):

docs/budgeting/categories.md
```diff
@@ -57,8 +57,17 @@ You can see in the budget that our categories have now been successfully merged
...
+## Managing Categories
+
+Category groups can be *expanded or collapsed* by clicking on the arrow to the left of the Category group name. All category groups can be expanded or collapsed at once by using the three-dot menu at the top of the category list.
+
+Categories that are not currently needed can be *hidden* by using the down arrow by each category name. An example is budgeting savings for a large cost (e.g. washing machine). After the purchase is made, the category can be hidden. Hidden categories still impact your budget. When needed, hidden categories can be viewed by clicking on the three dots at the top of the category list. 
+
+![](/img/categories/CategoryToggleHidden.png)
+
+Hidden categories, when shown, are displayed in a lower-contrast color.
```
matt-fidd, line 68: It might be worth highlighting that hidden categories can be unhidden (it should be intuitive, but then again, so should hiding categories)

```suggestion
Hidden categories, when shown, are displayed in a lower-contrast color. If you find that you now require a category that you have previously hidden, you can unhide it by using the down arrow next to the category name.
```

matt-fidd (changes requested):
Thanks for this! Just one small comment.

PR #402, opened 2024-07-10
Revamp of A tour of actual

RubenOlsen:
The current Tour was lacking for people curious about what Actual looks like, without going into "full teaching manual mode".

What I have done:

- Consistent screenshots
- Added some explanatory text and links to the actual manual pages

I am fully aware that there is a fine line between some of these pages and manual pages.

Anyway - we should now have something to show the YNABers and MINTers of the world :-)

Ideally, we should have a video introduction as the "tour" - but given how much change we have in the program, this would kill the poor guy/gal who wants to maintain such a video.

PS: I go on vacation on Saturday July 13th - so any changes I am not able to do after that time.


docs/tour/index.md
youngcw, line 1: I think this is too long and has included things that probably should be other places, and are in other places.  I would think just giving an intro to the tour and listing the tour sections would be better.

RubenOlsen: I agree with you on this. I just continued to build on the existing works (and looked at how other softwares did this).  

At some point in time some of us need to get together at discus what the tour should really look like.  

I'll short it down before my vacation - and continue working on the tour in 2 weeks time.  

Let's get it out (as it is so much better than what we had) - and improve when I get back 😊

RubenOlsen: I have now trimmed the starting page for the tour down to just the minimum.

We need to consider the following:

1. Our documentation is not read in any given order by our users. This means that some information is worth duplicating (as in how easy Actual is to get started with) in several places in our materials.

2. As I wrote when I opened the PR, I am a bit ambivalent about the Tour itself and how it is placed in the hierarchy. As it is now - it's not really documentation - and would you go to the system's documentation to get to "a tour"?  A tour is more "marketing material" than documentation (in my opinion). For the days I have worked on this, I have one question that has been going on in my mind:  _Would it make sense to have a top menu item called "A Tour of Actual" and have the tour pages in their own directory rather than part of the documentation?_

Anyway - unless there is other feedback - let's get this out there :-)

docs/tour/budget.md
```diff
@@ -1,5 +1,78 @@
...
+
+![](/static/img/a-tour-of-actual/tour-budget-calendar.png)
+
+Based on this, you can then choose which months to show:
+
+![](/static/img/a-tour-of-actual/tour-budget-calendar-choose.png)
+
+
+
+## The Top Section
```
youngcw, line 21: Could this be called "Budget Header" or "Month Header"

docs/tour/budget.md
```diff
@@ -1,5 +1,78 @@
...
+
+
+![](/static/img/a-tour-of-actual/tour-budget-top-expanded.png)
+
+When the top is minimized, you can still access the same functionality as when the top section is expanded.
+
+![](/static/img/a-tour-of-actual/tour-budget-top-minimized.png)
+
+
+## The Budget Detail Section
```
youngcw, line 41: "Budget Table"?

docs/tour/accounts.md
```diff
@@ -1,9 +1,67 @@
...
 
-![](/img/accounts/budget-account-register.png)
+This view lets you manage your transaction for an account. Consult the [Accounts & Transaction](/docs/accounts/) section in
+the manual for detailed information on how to work with this view.
 
-### Cleared/Uncleared Account Balance
+![](/static/img/a-tour-of-actual/tour-account-register-overview.png)
 
-When viewing your account in Actual, it is possible to show the Cleared & Uncleared totals of that account, to do this, hover over the account balance and an expando icon will appear, click that and your cleared and uncleared balances will appear.
+## The Account Register Top Section
```
youngcw, line 8: "Account Header" maybe?  I feel like calling so many things "section" sounds weird.

docs/tour/reports.md
```diff
@@ -0,0 +1,37 @@
+# The Reports View
```
youngcw, line 1: Can we call this just "Reports" since we are showing off the feature more than the page itself?

docs/tour/rules.md
```diff
@@ -0,0 +1,14 @@
+# The Rules View
```
youngcw, line 1: Same as reports.  Maybe "Transaction Rules"

docs/tour/schedules.md
```diff
@@ -1,5 +1,18 @@
-# Schedules
+# The Schedules View
```
youngcw, line 1: Maybe "Scheduled Transactions"

docs/tour/payees.md
```diff
@@ -0,0 +1,28 @@
+# The Payees View
```
youngcw, line 1: Maybe "Payee Management"

RubenOlsen:
Added the proposed changes, and a few others to reflect removing of "View" (very good catch by @youngcw).

Any changes other changes needs to wait 14 days due to vacation.

youngcw (approved):

PR #403, opened 2024-07-10
Add getAccountBalance() API documentation

psybers:
Docs for https://github.com/actualbudget/actual/pull/2930.

### DIFF
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -30,7 +30,8 @@ This is the documentation of all available API methods. The API has not been rel
 "updateAccount",
 "closeAccount",
 "reopenAccount",
-"deleteAccount"
+"deleteAccount",
+"getAccountBalance"
 ]} />
 
 <APIList title="Categories" sections={[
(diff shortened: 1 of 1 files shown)


RubenOlsen (approved):
Looking good to me.

docs/api/reference.md
```diff
@@ -303,6 +304,12 @@ Reopen a closed account.
 
 Delete an account.
 
+#### `getAccountBalance`
+
+<Method name="getAccountBalance" args={[{ name: 'id', type: 'id' }, { name: 'cutoff', type: 'Date?'}]} returns="Promise<number>" />
+
+Gets the balance for an account.  If a cutoff is given, it gives the account balance as of that date.  If no cutoff is given, it uses the current date as the cutoff.
```
youngcw, line 311: Does it actually use the current date or does it just use all transactions?

psybers: Current date.  Default behavior is 'give me the balance now' so future transactions (if they exist) should not be in there.

PR #423, opened 2024-08-04
Converts Norwegian Trumf Visa PDF invoice to an AB friendly CSV file

RubenOlsen:
Added link to converter from Trumf Visa PDF invoice to Actual Budget friendly CSV import file.


MatissJanis (approved):
Thanks!

PR #424, opened 2024-08-05
Update fly.io pricing information

Jonathan-Fang:
Fly.io no longer offers free cloud hosting; it is pay as you go with actual usage billed monthly. I calculated between Pikapods $1.4 per month and Fly.io's $1.5 per month and chose Pikapods since it's one click and it's cheaper. If there's anything wrong in my calculations, please let me know.


RubenOlsen (approved):

MatissJanis (approved):
Thanks

PR #425, opened 2024-08-08
Update report budget docs

psybers:
First, the way to enable it has changed so that is updated.

Second, the UI now calls it "tracking" and "envelope" budgeting, so the text and filenames were updated to reflect the new "tracking" name.


MatissJanis (approved):
Thanks!

PR #429, opened 2024-08-12
Reorganising the menu items

RubenOlsen:
1. Moved _Filtering_ from _Reports & Filters_ to _Accounts & Transactions_ as it's are more natural place to have the filter documentation.
2. Renamed _Reports & Filters_ to just _Reports_
3. Moved _Schedules_ out of _Budgeting_ to better reflect the sidebar items in the program.
4. Based on feedback from Michael Burnside on Discord, moved the envelope budgeting docs higher up in the menu. What he wrote on Discord: _I think the Envelope Budgeting doc should be moved up in the list. It's important to understand the method and the "Why" first, so that users can make sense of the tool and the "how". I know there's a link at the top of the Starting Fresh article, but people tend to go top to bottom in the contents (if they're reading at all)._
5. Moved _A Tour of Actual_ out of _Getting started_ as it's marketing and not getting started with the program.
6. Moved _Starting Fresh_ higher up to differentiate between the tech stuff and the why stuff.

**There is no change in content - just trying to make the menu a better place to navigate**


youngcw:
Misc thoughts:
* I like the new place for the tour.
* I think starting fresh should be before the tip&tricks.  Maybe tips&tricks and starting fresh could be moved down into the "Using Actual" section?
* moving filters is good.
* Maybe we should moved rules out of "Budgeting" and into the transactions section.  Or even combine filters and rules.  Those are very related ideas.
* Splitting out schedules is fine.  That page/section probably could use more into at some point.

RubenOlsen:
> Misc thoughts:
> 
> * I like the new place for the tour.

Thanks. How about also putting the tour on the top-page menu? 


> * I think starting fresh should be before the tip&tricks.  Maybe tips&tricks and starting fresh could be moved down into the "Using Actual" section?  I think that would make more sense than in the getting started section.

Tips and Tricks is after Starting Fresh?

image

I do agree that tips&tricks should be part of the "how to use" section of the menu.



> * moving filters is good.
> * Maybe we should moved rules out of "Budgeting" and into the transactions section.  Or even combine filters and rules.  Those are very related ideas.

Depending on the day I think about this, I do agree on this - or I do not agree with myself :-D

Conceptually, filters are ad hoc (I know you can save them) and are, I assume, mostly used with the account register, while rules are permanent. Because of this I belive we should have two separate documents. 

> * Splitting out schedules is fine.  That page/section probably could use more info at some point.

I want to re-organise the menu as follows. The aim is to have "day-to-day" information readily available, and information that is seldom used or one-off, like installing - should be less prominent.

- Introduction
  - Actual Budget Vision
  - A Tour of Actual
  -  Release notes (_alas, should it really be here and not somewhere else?_)
- Getting started
  - Roadmap
  - Envelope Budgeting
  - Migrating (_with the current sub pages_)
  - Installing and configuring Actual (_with current sub pages_)=
    - Configuring the Server _should be moved here_
    - Authenticating with HTTP Headers _should be moved here as it has to do with config/setup_)
- Using Actual
  - Tips & Tricks
  - On a mobile phone (_
  - Budgeting 
    - All documents belonging to how to technically use the budgeting section (Categories, Rules, etc)
    - _Subsection with documents that deal with theory on budgeting - name TBD_
      - Return and Reimbursements
      - Managing Credit Cards 
      - Carrying Debt on Credit Cards
      - Carrying Debt on Overdraft accounts (_not written - content to be more or less the same as Carrying Debt on CC - maybe even put these together)
      - Strategies for handling joint accounts (_I think this document should be extended so that the title is _Strategies for budgeting with a significant other_ (or something like that))
      - Restarting your budget _should be moved here_
    - Schedules
    - Accounts and Transactions
      - Filtering Transactions
      - Managing transfers
      - Rules
      - Manually importing transactions
      - Managing payees
      - Bulk actions
      - Automatic bank sync (_this is the Connecting your bank_ document)
      - I would like to put the _Identify and apply transfers historically_ here as it belongs under this main section
    - Reports (_with current sub pages_)
    - Backup & Restore (_with current sub pages_)
    - Experimental features (_with current sub pages_)
- Help and Support
  - The FAQ document
  - Troubleshooting
  - Community Projects
  - Contributing (_with the current sub pages_)

What do you think @youngcw?

youngcw:
That structure looks good.  I think the only things I would change would be 
* switch migrating and installing.  To me you would want to know how to install before migrating
* don't put tips and tricks at the top of the using section.  I would think that would go towards the bottom.
* I would put the section on mobile use in the installation section or in getting started, but maybe I don't know what your goal is for that section's content.  I had also thought that maybe we could add mobile use sections sprinkled throughout the docs to show both desktop and mobile use of certain features.

docs/getting-started/starting-fresh.md
youngcw, line 1: line 104.  That link doesn't work.

docs/reports/reports.md
youngcw, line 1: something is off about this reports section.  The sidebar selects both the "Reports" header and the "Reports" sub page.  The page also says reports and filters where the section is just about reports.

RubenOlsen: Fixed this and also reorganised the content a bit.

docs/getting-started/starting-fresh.md
RubenOlsen: Fixed all lins.

RubenOlsen:
@youngcw I think I have it all now :-)

youngcw (approved):

RubenOlsen:
That is not a nitpic IMO - that is a oversight from my part. I fixed it in the last commit. Now we have:

image

PR #446, opened 2024-09-03
Added info on how Pikapods updates

Jonathan-Fang:
Added information on how Pikapods updates for each monthly release, including how this works with desktop clients.


MatissJanis (approved):
Thanks

PR #451, opened 2024-09-15
Added Link to UnRAID SSL setup in /docs/install/index.md

RubenOlsen:


matt-fidd:
What do you think to reposting the guide in a gist or similar instead of linking to a discord thread, we've no way to patrol the thread to make sure it stays the way it is. I suppose that's the same with the other links too, but Discord feels a little more open.

youngcw (approved):

PR #462, opened 2024-10-04
[WIP] Update ynab4.md

mullermn:
Added in a Known issues section for YNAB4 imports to cover an issue with split transactions not receiving correct payee details.

Once this issue is resolved the documentation note can be removed:
https://github.com/actualbudget/actual/issues/3561

### DIFF
--- a/docs/migration/ynab4.md
+++ b/docs/migration/ynab4.md
@@ -5,3 +5,6 @@
 - All of your existing YNAB4 budgets should be listed. Select the one you want to import.
 - Click **Import**
 - That's it!
+
+## Known issues
+- Split transactions may import incorrectly; in YNAB4 a split transaction tracks a payee at the parent level, while in Actual the payee shown at parent level is joined from the payees of the splits. Because the splits do not have the correct payees they may not show up as expected in reports and filters. The fix for this is to use Settings / Show advanced settings / Repair split transactions. This will only need to be done once.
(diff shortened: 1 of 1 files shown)


docs/migration/ynab4.md
```diff
@@ -5,3 +5,6 @@
 - All of your existing YNAB4 budgets should be listed. Select the one you want to import.
 - Click **Import**
 - That's it!
+
+## Known issues
+- Split transactions may import incorrectly; in YNAB4 a split transaction tracks a payee at the parent level, while in Actual the payee shown at parent level is joined from the payees of the splits. Because the splits do not have the correct payees they may not show up as expected in reports and filters. The fix for this is to use Settings / Show advanced settings / Repair split transactions. This will only need to be done once.
```
RubenOlsen, line 10: Some small nit pickings:

```suggestion
- Split transactions may import incorrectly. In YNAB4, a split transaction tracks a payee at the parent level, while in Actual, the payee shown at the parent level is joined from the payees of the splits. Because the splits do not have the correct payees, they may not appear as expected in reports and filters. The fix is to use Settings / Show advanced settings / Repair split transactions. This will only need to be done once.
```

RubenOlsen (changes requested):
Looking good.

RubenOlsen (approved):
Looking good. Approved.

PR #467, opened 2024-10-09
add copy template

youngcw:
add blurb for the new copy template from actualbudget/actual#3617


RubenOlsen (approved):
Good to go!

RubenOlsen:
Question @youngcw  - is it only 12 months or do we also support _#template copy from 3 months ago _?

PR #482, opened 2024-11-04
docs: Explain running balance is shown only with descending date

ankitson:
I was a little confused that I couldn't see a running balance in my accounts until I realized that it's only shown in a specific sort order. 

I think it would actually be useful to be able to see the running balance no matter the sorted order, but until then maybe we should clarify in the documentation.

### DIFF
--- a/docs/getting-started/tips-tricks.md
+++ b/docs/getting-started/tips-tricks.md
@@ -60,7 +60,7 @@ This only controls the _maximum_ number of months. If the app is too small to re
 
 ## Show running balances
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/tips-tricks.md
```diff
@@ -65,8 +65,9 @@ A "running balance" is the balance of the account after every transaction over t
 To enable this:
 
 1. Click on an account
-2. Click on the 3 dots to show the actions menu
-3. Select "Show running balance"
+2. Make sure the transactions list is sorted by date in descending order - this option and the running balance are only shown in this case.
```
youngcw, line 68: Could this be added to the description above instead of in the steps here?  I think that would make more sense.

ankitson: Done!

PR #486, opened 2024-11-13
use built-in Docker init, remove tini

meonkeys:
We no longer need tini with recent versions of Docker.

See:

* https://docs.docker.com/reference/cli/docker/container/run/#init
* https://github.com/krallin/tini?#using-tini

Tested and seems to work as expected when `--init` is included in `docker run` command line or `init: true` is included in the Docker Compose service stanza.

Look for a correspoding change in the actual-server repo.


meonkeys:
see https://github.com/actualbudget/actual-server/pull/502

RubenOlsen:
Added the *Merge at release*

PR #487, opened 2024-11-15
Adding note that templates using schedules need to be on the same line.

redshiftltd:
My first edit, I tried setting up a schedule in a template and had to trial and error to get it to link. I think this is clearer.

### DIFF
--- a/docs/experimental/goal-templates.md
+++ b/docs/experimental/goal-templates.md
@@ -52,7 +52,7 @@ You are welcome to have other lines in your note, but the #template/#goal line m
 - Currency symbol is optional, `#template $50` and `#template 50` are the same.
 - Number formats that use comma for the decimal separator are not supported (eg, 123,45). You must use 123.45.
 - Thousands separators are not supported (eg, 1,234). You must use 1234.
-- {SCHEDULE NAME} is defined in the **Schedules** editor. **Take great care to copy across these schedule names EXACTLY**, without braces.
+- {SCHEDULE NAME} is defined in the **Schedules** editor. **Take great care to copy across these schedule names EXACTLY**, without braces. Templates with schedules must be on one line.
 - By default templates do not consider available funds when being applied. Use template priorities to not budget more than is available.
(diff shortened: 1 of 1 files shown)


youngcw:
what do you mean by "on one line"?

youngcw:
This would apply to all the templates to show up as multiple lines in the table. Maybe this should be added somewhere else since its not specific to the schedule template.

redshiftltd:
Yes, I agree there. I didn’t want to push for too much so that’s why I put it as a footnote.  From: youngcw ***@***.***>Date: Friday, November 15, 2024 at 17:06To: actualbudget/docs ***@***.***>Cc: redshiftltd ***@***.***>, Author ***@***.***>Subject: Re: [actualbudget/docs] [WIP] Adding note that templates using schedules need to be on the same line. (PR #487)This would apply to all the templates to show up as multiple lines in the table. Maybe this should be added somewhere else since its not specific to the schedule template.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

youngcw:
Your wording is say that all templates must be on the same line, not just that a given template must be a single line.  Could you update it to be clear its not just schedules and that its referring to a single template and not all templates?

RubenOlsen:
@redshiftltd, any comment on the feedback from @youngcw?

redshiftltd:
@youngcw I added the note as a new line and to me it looks clearer. Please take a look and tell me if it makes sense to you.

RubenOlsen:
@redshiftltd, when you are happy with your changes, remove the [WIP] from the summary/subject.

redshiftltd:
@RubenOlsen Thank you, done!

youngcw:
@redshiftltd I recently rewrote the template docs page.  I just added an additional PR that adds a note that I think covers this plus makes it clear that it applies to all templates not just schedules.  Ill close this PR, but feel free to comment on #611 if you think it should be more clear.

PR #497, opened 2024-11-26
adds hint for missing date formats

ewatch:
If the date column is not configured correctly the date format will not show up if it's not one of the preconfigured ones.
I think that can help in the documentation.


RubenOlsen (commented):
Looking good to me. When you are ready @ewatch, plase remove the [WIP] from the summary and the formal review will start.

RubenOlsen (approved):
Looking good - will merge.

PR #521, opened 2024-11-27
fix some spelling issues, standardis(z)e to US English

matt-fidd:
Quick whip through to try and standardis(z)e some language, looks like the most common English in the docs is US English so I chose that one, but if we'd prefer the Queen's then that's okay too.


RubenOlsen (approved):

RubenOlsen (commented):
I notice the americanizations - and in general, and in the future, I do not think we should change those forms as long as there are no typos and there is consistency within a specific document. 

Some people, like myself, are not from the US and I (try to) use UK English spelling forms :-)

PR #523, opened 2024-12-02
Update reverse-proxies.md

spoon1331:
After fighting with this for some time, it seems that the trailing slash in apache config causes budget to think that the server is offline. Removed.


RubenOlsen:
When you are ready to have a review of your changes, just remove the [WIP] from the summary.

RubenOlsen (approved):
Looking good and has been technically verified - will merge.

PR #543, opened 2024-12-03
Update restore.md

bpaulien:
title case, punctuation


RubenOlsen (approved):
Looking good - will merge.

PR #551, opened 2024-12-06
Update goal-templates.md

bpaulien:
remove currency symbols, add note explaining how they're used, tried to standardize how template vs #template / goal vs #goal are used, some edits for clarity,


RubenOlsen (approved):
Very good work - will merge.

PR #560, opened 2024-12-08
Update reconciliation instructions and images to reflect current UI

adivardhan1000:
When I setup my actual account, I noticed that there was no reconciliation button as described in documentation. Found out after hunting, that the UI has changed resulting in this issue. I have created this PR to update that.

Let me know if the language or images need to be changed.

Caused because of: https://github.com/actualbudget/actual/pull/3684


matt-fidd (approved):
Great, thank you!

PR #586, opened 2024-12-24
Add docs about releasing translations

jfdoming:


RubenOlsen:
This is looking good to me - I added the **Merge at release** label.

jfdoming:
> This is looking good to me - I added the **Merge at release** label.

Ah thanks for taking a look! These docs are actually pretty stale, I'm now planning on taking a different approach after discussing with the others (Git submodules are really un-ergonomic 😞). Will close the PR for now assuming I won't take this approach (but we can reopen if that changes)

PR #607, opened 2025-01-09
Update instructions for enabling Tracking Budget

KarboniteKream:
This PR updates the instructions for enabling Tracking Budget. The feature is no longer experimental, but the documentation still treats it as such.


KarboniteKream:
Confirmed that the deploy preview works as expected.

PR #612, opened 2025-01-19
Add note on how to update client version

Kennedy242:
From the docs it isn't clear how to update
the client when using docker. This has been
asked a few times on the discord channel.

### DIFF
--- a/docs/install/docker.md
+++ b/docs/install/docker.md
@@ -45,6 +45,8 @@ You can optionally configure the container using environment variables — see t
 $ docker-compose pull && docker-compose up -d
 ```
 
+> Note: After updating the server container, the client will automatically be updated. A manual page hard reload may be required to see the changes.
+
 ## Launch container using docker command
 
 Pre-requisites: Docker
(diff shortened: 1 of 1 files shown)


docs/install/docker.md
```diff
@@ -45,6 +45,8 @@ You can optionally configure the container using environment variables — see t
 $ docker-compose pull && docker-compose up -d
 ```
 
+> Note: After updating the server container, the client will automatically be updated. A manual page hard reload may be required to see the changes.
```
jfdoming, line 48: A hard reload should no longer be required with the latest client version, it will instead prompt you to update in a notification. Were you seeing different behaviour?

Kennedy242: Oh okay good to know. I just updated server version from v24.11.0 to v25.1.0 and needed to close my firefox browser and then the client was updated. A page refresh wasn't working to update the client.

jfdoming: Ah yeah, 24.11 is the last version that didn't have that feature. Starting with the next updates it should work

Kennedy242: Great thanks for the info! I'll go ahead and close this PR.

PR #641, opened 2025-02-13
Grammar correction to starting-fresh.md page

davidkaufman:
s/like bill and groceries/like bills and groceries/


matt-fidd (approved):
Thanks!

PR #645, opened 2025-02-15
Upating Local install docs for people on a clean clone

MikesGlitch:


matt-fidd (approved):
Seems like the best solution of the three, saves a fair chunk of data transfer and space.

PR #648, opened 2025-02-24
add gcp cloud run install option

daniefdz:
I created a repository with steps to deploy Actual Budget as a Cloud Run service. I believe this is a better way to deploy in GCP and it's less complicated.


matt-fidd (approved):
Thanks for this!

PR #653, opened 2025-03-01
Add docs on data to move

jfdoming:


docs/actual-server-repo-move.md
```diff
@@ -41,7 +41,8 @@ The reasons for this change are as follows:
   ```
   packages/sync-server
   ```
-  5. Run the server with:
+  5. Copy over your data (`user-files`, `server-files`, and `.migrate`) into the same `packages/sync-server` directory.
```
youngcw, line 44: Maybe just say 'into' instead of 'into the same'.  That seems less confusing

youngcw: And probably point out where the user-files, etc are at in the first place

jfdoming: Good point

youngcw (approved):

PR #673, opened 2025-04-02
🔖 (25.4.0)

matt-fidd:
- web https://github.com/actualbudget/actual/pull/4728
- docs https://github.com/actualbudget/docs/pull/673


jfdoming (approved):

youngcw (approved):
I think we should call out the embedded sync server somewhere, but it doesn't have to be here.

matt-fidd:
> I think we should call out the embedded sync server somewhere, but it doesn't have to be here.

Sorry, I misinterpreted this as the workspace reference to web for sync-server - I hadn't even seen Mike's change go in this month! I agree it should be called out, I'll edit the docs now.

PR #692, opened 2025-05-12
change minimum version of Node.js to version 20

matt-fidd:
I think I got all instances where we specify a version, there seem to only be two places in the docs.

We'll also call it out in the release notes.

https://github.com/actualbudget/actual/pull/4978


RubenOlsen (approved):
Approved.

PR #706, opened 2025-05-24
List Arch AUR packages

pogman-code:
There was one package in the doc (actual-appimage) but there is more actually :D  
2 Desktop clients based on the AppImage and a server based on the @actual-app/sync-server NPM package.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -4,6 +4,7 @@ Activo
 AESUDEF
 ALZEY
 Anglais
+AUR
 Authentik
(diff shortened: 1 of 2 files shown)


pogman-code:
I have added details for each packages, based on the response of the maintainer of actual-appimage package (https://discord.com/channels/937901803608096828/1027831463103696928/1375945898994958367)

RubenOlsen:
I see @youngcw's comment and while I agree I'd rather see the list in alphabetical order. If we truely consider this to be "expert options" we may think about a reorg of that section.

PR #707, opened 2025-05-26
Fix ambiguous examples in goal-templates page

Mansarde:
The last two examples for the goals have created confusion because the template and goals line shows in one line:
https://discord.com/channels/937901803608096828/1372691110614274170/1376211790639136779

This PR makes sure that any `#template` and `#goal` lines that are meant to go together are displayed on top of each other instead of inline.

It also makes sure that all uses of `#template` show with inline code-formatting for overall conistency.
(Let me know if you'd rather have this in a separate PR and I'll splice out that commit.)

### DIFF
--- a/docs/experimental/goal-templates.md
+++ b/docs/experimental/goal-templates.md
@@ -109,14 +109,14 @@ Templates can be given a priority flag to change the order that the templates ge
     - **Expected budgeted amount is 200** 
     **Expected maximum category balance is 200**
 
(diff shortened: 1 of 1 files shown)


Mansarde:
Hmm, it looks like the deploy-preview doesn't like BR's as a closing tag.
Will change that and see if that fixes it.

Mansarde:
Hold up, still need to fix something.

PR #729, opened 2025-06-22
Documentation for Schedule API based on CRUD.

karimkodera:
This documentation is based on code in PR to be assign. The docs suggests a structure for a schedule entity and recur confirm. It also provides create, get, update and delete functions. 2 unrelated functions have been added to get server version to get the ID of schedules, categories, payees or accounts by name.


matt-fidd (approved):
LGTM!

PR #741, opened 2025-07-03
chore: added link to correct gocardless service so people sign up to the right gocardless service

dariusrosendahl:
After registering for a completely unnecessary GoCardless main account, I found out you actually need GoCardless Bank Account Data acount. Lets point other new users in the right direction.

### DIFF
--- a/docs/advanced/bank-sync.md
+++ b/docs/advanced/bank-sync.md
@@ -13,7 +13,7 @@ Here are a couple of considerations to know about before making the decision to 
 
 ## Supported Providers
 
-* GoCardless (European Banks)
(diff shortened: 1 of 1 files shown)


youngcw:
That link is confusing, most of the buttons on that page go to the regular gocardless, not the personal use one.  Also, there is a link the in the gocardless docs page already.

dariusrosendahl:
> That link is confusing, most of the buttons on that page go to the regular gocardless, not the personal use one. Also, there is a link the in the gocardless docs page already.

Sorry, agreed. The whole GoCardless situation was confusing. That's why I wanted to add a link to their correct service.

I mend to add this one: https://bankaccountdata.gocardless.com, I just had to google it but I found the link in the documentation. Changed it to that!

PR #750, opened 2025-07-18
Fix Bank Sync in docs and add new picture

deathblade666:


MikesGlitch (changes requested):
Nice! 

Minor thing, can you change the image to be in light theme? Just to keep it the same as most of the other images across the docs.

deathblade666:
> Nice!
> 
> Minor thing, can you change the image to be in light theme? Just to keep it the same as most of the other images across the docs.

Done

MikesGlitch (approved):
Thanks!

PR #754, opened 2025-07-21
Better readability and fixes

Pheggas:
I changed two docs pages - multi user and oauth (opeid provider).

There were some minor inconsistencies and errors. I also fixed the redirect URI from `<domain>/path/openid/callback` to `<domain>/openid/callback` as the first one was incorrect for my keycloak instance.

### DIFF
--- a/docs/config/multi-user.md
+++ b/docs/config/multi-user.md
@@ -1,6 +1,10 @@
 # Managing Multi-User Support
 
-This feature requires you have setup an [OpenID Provider](oauth-auth). The usernames will be fetched from the provider.
+:::caution
+
+This feature requires you have set up an [OpenID Provider](oauth-auth). The usernames will be fetched from the provider.
+
+:::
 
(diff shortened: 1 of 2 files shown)


docs/config/multi-user.md
```diff
@@ -1,6 +1,10 @@
 # Managing Multi-User Support
 
-This feature requires you have setup an [OpenID Provider](oauth-auth). The usernames will be fetched from the provider.
+:::caution
+
+This feature requires you have set up an [OpenID Provider](oauth-auth). The usernames will be fetched from the provider.
```
RubenOlsen, line 5: We operate with absolute paths. 

```suggestion
This feature requires you to have set up an [OpenID Provider](/docs/config/oauth-auth). The usernames which people will log into your Actual instance will be fetched from the provider.
```

RubenOlsen (changes requested):

PR #768, opened 2025-08-07
Upgrade docusaurus

MikesGlitch:


youngcw:
Could we make some sort of lint setup to prevent these small syntax issues going forward?

PR #771, opened 2025-08-09
Change search plugin to a more accurate one

MikesGlitch:
The search tool we use has always been a bit clunky for me. There's no keyboard shortcut support, it doesn't look great, doesn't support react 19, and I had to write a theme sync function for it to work properly with Docusaurus V3. 

The package I've upgraded here fixes all those issues. I've also updated all the packages to the latest version, and aligned the node support with the Actual repo.

**Old (very inaccurate search):**

image

**New:** 

image

It also comes with a fancy new search page when you click _See all results_


matt-fidd:
Sorry, they put that button in the same place as the "merge with latest master" button

RubenOlsen (approved):

PR #802, opened 2025-09-22
document api quiet mode

matt-fidd:
For https://github.com/actualbudget/actual/pull/5762

### DIFF
--- a/docs/api/types.js
+++ b/docs/api/types.js
@@ -65,6 +65,15 @@ export let objects = {
         </span>
       ),
     },
+    {
+      name: 'quiet',
+      type: 'boolean',
+      description: (
+        <span>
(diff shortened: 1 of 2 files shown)


docs/api/types.js
```diff
@@ -65,6 +65,15 @@ export let objects = {
         </span>
       ),
     },
+    {
+      name: 'quiet',
```
MatissJanis, line 69: Should be "verbose"

matt-fidd: Good catch, doh.

https://github.com/actualbudget/docs/pull/802/commits/a5e1cf0a68ab0a1afe59e977f405586cda45404d

MatissJanis (approved):

PR #803, opened 2025-09-22
rule-templating: clarify meaning of sub prefix in date functions

MrGreenTea:
I was confused at first and was thinking of "substitute". So I thought it might be a nice addition to the docs.

Additionally I cleaned up the markdown tables a bit and added a link to the patterns available for date `format`


matt-fidd (changes requested):
Thanks for this! Could you make sure the capitalisation is consistent please?

matt-fidd (approved):
Brilliant, thank you!

